PR #16, opened 2021-05-21
Multiple external disk mounting capability

elmerenges:
I just tweaked your code, now it is possible to mount multiple external shares.
It is backward compatible with the previous versions.

### DIFF
--- a/repository.json
+++ b/repository.json
@@ -1,5 +1,5 @@
 {
-  "name": "Hass.io Add-ons by alexbelgium2",
+  "name": "Hass.io Add-ons by alexbelgium",
   "url": "https://github.com/alexbelgium/hassio-addons",
   "maintainer": "alexbelgium"
 }
--- a/repository.json
+++ b/repository.json
@@ -1,5 +1,5 @@
 {
-  "name": "Hass.io Add-ons by alexbelgium",
+  "name": "Hass.io Add-ons by alexbelgium2",
   "url": "https://github.com/alexbelgium/hassio-addons",
   "maintainer": "alexbelgium"
 }
--- a/sonarr/config.json
+++ b/sonarr/config.json
@@ -26,7 +26,7 @@
     "PUID": 0,
     "PGID": 0,
     "localdisks": ["sda1"],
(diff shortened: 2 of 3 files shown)


alexbelgium:
Thanks! I'll merge this and update the version to allow people to update

alexbelgium:
I've also added a code to mount smbv1 shares : it is an optional boolean in config.json. If checked, then in 50-mounts the code is added to cifs mounts.

Thanks again for helping making those addons better!

alexbelgium:
Hi, I have made some changes to the code to make it more compatible following an issue with a user. Due to this, there are two breaking changes : multiple network disks must be separated by a "," and they are mounted to a folder with the name of the external share.

brewcrawler:
It's a reasonable change. I was super lazy with my PR, just wanted it to work somehow. This version is much nicer.

elmerenges:
It's a reasonable change. I was super lazy with my PR, just wanted it to work somehow. This version is much nicer.

alexbelgium:
No I really appreciate your work and that you took the time for the PR !

Actually someone had trouble mounting his smb drive, and in the other code I had implemented a sequential testing of all cifs versions and securities until it works... And this allowed him to mount the smb drive.

But please if there was another option that I removed when changing the code please tell me and I'll adapt

Have a nice day

elmerenges:
I think the current approach is better then the previous.
I updated both Radarr and Sonarr, and the additional tinkering with my previous remote mount settings took about 5-10 minute, so I am happy :)

Thank you for the addons, they saved me from some headache.

alexbelgium:
btw I've added a new addon from the same lineage, prowlarr. As I understand, it is an help program for Radarr and Sonarr "Prowlarr is a indexer manager/proxy built on the popular arr .net/reactjs base stack to integrate with your various PVR apps. Prowlarr supports both Torrent Trackers and Usenet Indexers. It integrates seamlessly with Sonarr, Radarr, Lidarr, and Readarr offering complete management of your indexers with no per app Indexer setup required (we do it all)."

PR #78, opened 2021-11-09
Ingress support

ananthb:
Changes:
- Ingress support
- Host networking to support mDNS for chromecast
- Expose /dev/vchiq for hw accelerated transcoding on raspberry pis.


alexbelgium:
Hi, thanks for the proposal! However ingress support is more complex to set. You need to add nginx to the addon app, map the ports, then (biggest issue) map the websocket if there are any...

Regarding mDNS, can't it be set with the addon apparmor instead of broadcasting to the whole host network? As this can interfere also with other addons.

What do you think?

alexbelgium:
Have you tested all changes you propose?

The way to do it is merge it in your own forked repo, install it on your HA instance, install the addon and check that your changes perform as expected.

Thanks!

ananthb:
I tried it out, and it doesn't work. I set ingress_entry to /web and to /web/index.html but I'm still getting a 404. I'm going to look into this some more to see what needs to happen. I also saw something in the logs about websockets, so I'll probably have to do something about that too.

mDNS unfortunately doesn't work reliably with docker's bridge networking. If you know any other addons that have successfully done this without host network mode, I'll gladly copy that over.

alexbelgium:
For video errors I've updated the config.json and apparmor.txt to allow thoses devices. Hopefully this will solve it. It's perhaps a good idea that we already merge it, then think a bit more for the two others what would be the impact on other users.

For mDNS I'll check... Indeed there shouldn't be any conflict as port 8096 is not a usual default port, except for people using emby but normally people wouldn't have both emby and jellyfin running in parallel...

For ingress, are you sure it is possible to implement without nginx? On my other addons I've usually implemented it by putting a layer of nginx that redirected the ingress port, such as this example here that forwards both http and websocket : https://github.com/alexbelgium/hassio-addons/blob/master/joal/rootfs/etc/nginx/servers/ingress.conf

alexbelgium:
Hi, are you still interested in this PR ? Not sure why but it seems to try to merge all files instead of just the ones you modified, which creates a huge mess

ananthb:
Yeah I am. Didn't have much time to look at it this week. I'll probably have a new PR with the changes I'm working on soon.

alexbelgium:
Thanks, no stress ;)
I created a second version of the jellyfin addon (named test) to check ingress with nginx but couldn't make it work either...

alexbelgium:
Hi, you're cancelling? Is it not working or are you not interested anymore?

ananthb:
I'm definitely interested! Trying to streamline my open PRs right now. I'll open a new PR when I've progressed some more on this.

PR #80, opened 2021-11-11
Official Portainer Add-On Deprecated

EpicLPer:
The official Portainer Add-On has been deprecated by Franck, the URL should in general be changed to this add-on page instead for ease of access.


alexbelgium:
Thanks!

PR #130, opened 2021-12-26
Made a lot clearer why this addon exists ;)

bruvv:


alexbelgium:
Thanks! Much clearer indeed. Don't hesitate to propose any other improvements! I don't have tons of time to put on hassio addons so all help is welcome :)

PR #132, opened 2021-12-26
Change config file location fix EOL and fixed some extension

bruvv:
Make sure to use a special folder for the config files.


alexbelgium:
Great idea, thanks! 

To avoid a breaking change what we can do is check the initial location to see if a file exist, then move it to the new one.

I've implemented in Readarr a code to allow to select the config location dynamically from the addon options if you are also interested (it was a request from an user)

bruvv:
it's already in the code. it checks if the folder exists and moves the stuff away.

bruvv:
@alexbelgium I pushed some new commits and all should be good now.

alexbelgium:
Thanks for your work! I've started looking at it.

I had three main elements/questions/reflections :
- Regarding the config location, we also need to modify the dockerfile to dynamically modify the config locations from upstream
- Why did you include blank lines at the 
- Regarding the files rename (.sh added) the fact of keeping them as such was sometime to overwrite the file coming from the upstream image ; we'll have to check to make sure that there is not some leftovers from the upstream image that duplicate functions (or lead to an addon failure). Usually I tried to avoid overwriting file, but the nextcloud addon for example was initially written by someone else and I think he based his way of working on overwriting upstream files

I'll continue looking at it, great to see your interest in making the addons better! thanks!

alexbelgium:
Thanks for your work! I've started looking at it.

I had three main elements/questions/reflections :
- Regarding the config location, we also need to modify the dockerfile to dynamically modify the config locations from upstream
- Why did you include blank lines at the start of some sh files ? is it a best practice? I'm not an expert but thought that it might not be conform to sh standards that requires the #! on the first line?
- Regarding the files rename (.sh added) the fact of keeping them as such was sometime to overwrite the file coming from the upstream image ; we'll have to check to make sure that there is not some leftovers from the upstream image that duplicate functions (or lead to an addon failure). Usually I tried to avoid overwriting file, but the nextcloud addon for example was initially written by someone else and I think he based his way of working on overwriting upstream files

I'll continue looking at it, great to see your interest in making the addons better! thanks!

bruvv:
Sorry for the delay of reply:

- Regarding the config location, we also need to modify the dockerfile to dynamically modify the config locations from upstream
check will double check this

- Why did you include blank lines at the start of some sh files ? is it a best practice? I'm not an expert but thought that it might not be conform to sh standards that requires the #! on the first line?
This is something my exention did in VScode by modifing all the end of line sequence. This was set to windows CRLF and since we are doing a lot on linux you want LF. And I guess some files got new lines which indeed can be ingnored.

- Regarding the files rename (.sh added) the fact of keeping them as such was sometime to overwrite the file coming from the upstream image ; we'll have to check to make sure that there is not some leftovers from the upstream image that duplicate functions (or lead to an addon failure). Usually I tried to avoid overwriting file, but the nextcloud addon for example was initially written by someone else and I think he based his way of working on overwriting upstream files
This is just a really bad way of doing it, by removing the exentions is not really solving the issue. Can you name the problematic addons? I will look into this more.

nextcloud/Dockerfile
```diff
@@ -26,7 +26,10 @@ FROM ${BUILD_FROM}
 ENV NEXTCLOUD_PATH="/data/config/www/nextcloud"
 
 RUN \
-    chmod +x /defaults/nextcloud-perms.sh \
+    # Remove upstream files
```
alexbelgium, line 29: avoids duplication with the files you renamed

bruvv: nice one :)

bruvv:
@alexbelgium please please change to LF: https://stackoverflow.com/questions/2517190/how-do-i-force-git-to-use-lf-instead-of-crlf-under-windows

alexbelgium (approved):

alexbelgium:
Thanks, I'll merge it! I've added the comments in changelog and updated the dockerfiles.

alexbelgium:
ah ah we were looking at the files in parallel! I just saw your comments and you have really good points.

- Dockerfile
Hopefully it's done! My addons coding have evolved with time and my knowledge progression. It would probably be better to grep & sed (or just sed) recursively for all /config instances instead of replacing files one by one as I did at the beginning. But I don't have much time to test and I'm only using a very few out of all the addons

- "CRLF to LF" 
-> ah indeed this has bugged me so often... I don't understand my vs code is supposedly configured to use LF but it always reverts to CRLF... I'll read your documentation to see how to make this better! Periodically I just run dos2unix in git bash to re-convert all files but it really isn't efficient ;-)

- Overwriting files
"This is just a really bad way of doing it in my opinion" -> totally agree :-) Actually I usually try to sed files from the dockerfile in the "modify upstream image" as it is much more sustainable : if the maintainer updates the upstream image it will then keep things working! It is mainly for the nextcloud addon that files are duplicated, as I initially forked the addon from another user (haberda) and never took the time to rewrite his logic as it would take much time, and also I would risk breaking things for some users... As I see you are coding-inclined, don't hesitate to propose any other improvements to my addons! I don't have tons of time and I'm not a programmer by myself - at the basis I'm a bioingenier that self-taught everything in the past year, so things take me much longer with trials & errors.

alexbelgium:
One thing I was thinking about doing is using templates. I've done that for example to install dependencies, with a script that is downloaded during docker build, checks the scripts for commands used, and download dependencies with apk or apt-get depending on the system used.

I try to standardize as much as possible my scripts, so that when I update one (such as 92-smb_mounts.sh), I use copy & replace to replace the scripts in all my addons.

I was wondering if I couldn't use dynamic download, to download scripts from a single source during build and avoid code duplication... That's one thing I will look at in the future. Your opinion is welcome on this if you see a more efficient way to do it :-)

alexbelgium:
> @alexbelgium please please change to LF: https://stackoverflow.com/questions/2517190/how-do-i-force-git-to-use-lf-instead-of-crlf-under-windows

Thanks! Hopefully it worked :-)

bruvv:
Awesome Alex! Good work so far :) I am not a developer myself either but working in IT. I will see if I can do some more stuff for your awesome work so far!

alexbelgium:
Thanks!

PR #133, opened 2021-12-30
Allow changing the data directory of nextcloud

marciogranzotto:
I wanted to change my data directory of NextCloud to point to my external HDD (mounted as /media/mydrive)
So I added this option on the config


alexbelgium:
Great idea, and thanks for having thought to update readme and everything

Issue #135, opened 2022-01-01
Jackett NAS lost all its configuration


almico:
All of a sudden, one hour ago, all indexers disappeared. Even the API KEY changed.
This is what I see in the logs. HA didn't restart. Only Jackett NAS, somehow, restarted.

```
[cont-init.d] 00-aaa_dockerfile_backup.sh: executing... 
[cont-init.d] 00-aaa_dockerfile_backup.sh: exited 0.
[cont-init.d] 00-banner.sh: executing... 
-----------------------------------------------------------
 Add-on: Jackett NAS
 translates queries from apps (Sonarr, SickRage, CouchPotato, Mylar, etc) into tracker-site-specific http queries, parses the html response, then sends results back to the requesting software
-----------------------------------------------------------
 Add-on version: 0.20.216
 You are running the latest version of this add-on.
 System: Home Assistant OS 7.0  (aarch64 / raspberrypi4-64)
 Home Assistant Core: 2021.12.7
 Home Assistant Supervisor: 2021.12.2
-----------------------------------------------------------
 Please, share the above information when looking for help
 or support in, e.g., GitHub, forums
 https://github.com/alexbelgium/hassio-addons
-----------------------------------------------------------
[cont-init.d] 00-banner.sh: exited 0.
[cont-init.d] 01-envfile: executing... 
[cont-init.d] 01-envfile: exited 0.
[cont-init.d] 01-migrations: executing... 
[migrations] started
[migrations] no migrations found
[cont-init.d] 01-migrations: exited 0.
[cont-init.d] 10-adduser: executing... 
-------------------------------------
          _         ()
         | |  ___   _    __
         | | / __| | |  /  \ 
         | | \__ \ | | | () |
         |_| |___/ |_|  \__/
Brought to you by linuxserver.io
-------------------------------------
To support LSIO projects visit:
https://www.linuxserver.io/donate/
-------------------------------------
GID/UID
-------------------------------------
User uid:    0
User gid:    0
-------------------------------------
[cont-init.d] 10-adduser: exited 0.
[cont-init.d] 20-folders.sh: executing... 
Moving to new location /config/addons_config/Jackett
[cont-init.d] 20-folders.sh: exited 0.
[cont-init.d] 30-config: executing... 
[cont-init.d] 30-config: exited 0.
[cont-init.d] 90-custom-folders: executing... 
[cont-init.d] 90-custom-folders: exited 0.
[cont-init.d] 92-local_mounts.sh: executing... 
[...]

PR #136, opened 2022-01-01
Fixed jackett (now proper merge...)

bruvv:

PR #139, opened 2022-01-01
Updated whoogle + jackett + other stuff

bruvv:
- Whoogle is now better configurable.
- Jackett fixed errors in the new config folder (https://github.com/alexbelgium/hassio-addons/issues/135) and streamlined the version
- Made some cosmetic changes to the files
- Added default github files, hopefully changing the EOL for good
- Added gitignore
- Fixed mulitple SC2163 and SC2068
- imgbot did his work and made all the images around 70% smaller while keeping the right quality.


bruvv:
Argh I see it also merged the main branch. Will close this: https://github.com/alexbelgium/hassio-addons/pull/136 and keep this as main pull request.

alexbelgium:
Is this ready should I merge it? Thanks!

bruvv:
Wait 1 sec, made a small error.

bruvv:
@alexbelgium save to merge!

addons_updater/rootfs/scripts/00-aaa_dockerfile_backup.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/with-contenv bashio
```
alexbelgium: Hi, you can't use bashio here - the point is to install it if the dockerfile build did not work. It is needed in context such as lock builds on supervised installations where curl doesn't seem to work. This was tested with several users of my addon and this was the only working solution to allow local builds on supervised installs. But then it requires #!/bin/bash

bruvv: that is not good, sorry will change it all asap!

bruvv:
fixed sorry @alexbelgium

alexbelgium:
Hi, I don't know if you received a notification as they were above your last comment, but there were too last questions above. The main one is relative to the code structure to export a variable as ENV, I thought the initial code was valid (and at least it worked) and wanted to double check your proposal with you. Thanks!

bruvv:
I cannot find those comments. :( as far as I am awair this can be merged now.

alexbelgium:
The review comments appear above in the thread on my Web browser : I opened a review on the files modified and those comments are marked as pending 
- hassio-addons/zzz_templates/automatic_packages.sh
Sorry this needs to be bash too - this is the script called by 00-dockerfile_backup to install bashio (lines 157+)
- /00-global_var.sh
Are you sure that it is export line and not $line?
As the actual content of $line is something like VARIABLE=value, si I think we need to expand the variable line for the true variable VARIABLE to be exported. Do you see, what I mean?

again thanks very much for the additions and improvments!

bruvv:
Fixed the bin bash stuff.
About the export line you can read more on this here:
https://github.com/koalaman/shellcheck/wiki/SC2163

alexbelgium:
> Fixed the bin bash stuff. About the export line you can read more on this here: https://github.com/koalaman/shellcheck/wiki/SC2163

Hi, but this is my point : I do not want to export the variable line, but the variable that is in text inside.

line contains "VAR=value", so once $line is expanded, what the bash code sees is ```export VAR=value```

Example:
line="VAR=key"
export $line is therefore equivalent to export VAR=key

Do you see the rational?

bruvv:
Ah I see! I'm not home ATM so I will do this later tonight.

alexbelgium:
BTW you changed also the 00-global_var.sh script from bashio to bash but that one was correct :) as bashio is called within the script (bashio::config.false)

alexbelgium:
Actually there is only two scirpts that can't accept the bashio shebang : 00_dockerfile-something, that is used to install elements when the dockerfile install failed for supervised installs, and the automatic_packages.sh that is the only that actually installs bashio :) it can be a bit confusing but that is the way I standardized the addons logic :)

bruvv:
Ah yes pretty new to the whole basshio. So what you suggest to change for the merge ?

alexbelgium:
> Ah yes pretty new to the whole basshio. So what you suggest to change for the merge ?

I think everything is fine except for export $line and to revert to binbash for global var :) I'm not at home but that's the only remaing things I was seeing :)

bruvv:
Ah check I've already reverted the bash stuff in my last commits. Can you check them? I only have to change the line variable.

alexbelgium:
Looks great!

alexbelgium:
I've corrected the last few bashio and $line, let's merge!

Thanks again for all the proposals and improvements!

bruvv:
Awesome thanks for the commits :)

PR #140, opened 2022-01-03
Changed new upstream versions

bruvv:
Also added missing changelogs from the past few comments :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,0 @@
-repository.json
--- a/bazarr/CHANGELOG.md
+++ b/bazarr/CHANGELOG.md
@@ -1,5 +1,6 @@
-## 1.0.2-1
-- Fixed local config
+## v1.0.2-ls140
+- Fixed local config-
+- updated to v1.0.2-ls140
 
 ## 1.0.2 (01-01-2022)
 - Update to latest version from linuxserver/docker-bazarr
--- a/bazarr/CHANGELOG.md
+++ b/bazarr/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## v1.0.2-ls140
-- Fixed local config-
+- Fixed local config
 - updated to v1.0.2-ls140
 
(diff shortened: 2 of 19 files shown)


alexbelgium:
Hi, I'm outside for the moment and not near a usable github but actually the "upstream" code is automatically updated every Monday, Wednesday and Friday according to the addons_updated addon that runs on my rpi!

The logic is that the addon checks if the tag "upstream" is aligned with the latest release on github (or dockerfile). It does not look the tag "version". If there is a new release, it updates the tag "upstream" in 3 files if relevant : config.json, build.json and Dockerfile, and aligns the tag "version" with the new upstream. This allows to have a dockerfile for example with : ```BUILD_FROM name/name:$UPSTREAM```

If you update it manually, you need to make sure that all instances are changed, as the addons_updater works with sed to replace the tag. So if somewhere the tag upstream does not corresponds to the one in config.json, it will not be updated by my automatic script.

However, I usually link the docker build to "latest". This allows people to always have the latest version, even if my script did not update the version. This avoids issues if for example, the dockerhub image upstream is not named the same way as the github releases.

Hope it clarifies :)

I'll look at all your commits a bit later today! Thanks for the commits!

alexbelgium:
In the same way, my addons_updater script does not take into account the part after -ls in the upstream versions. Only the part to the left corresponds to the app version, the number after - ls corresponds to improvements made by the linuxserver team that manages some upstream image (for example, an automatic dependence update - useless for us as we are building from the docker image)

alexbelgium:
Thanks!

bruvv:
awesome didnt know that! thanks for explaining. How is the changelog populated? Is that also done by the addon? It should include the very specific version number.

alexbelgium:
Yes :) actually there is a high probability that all files realign this night when then addon runs, we'll see tomorrow. It works pretty well :)

PR #143, opened 2022-01-03
removed date from the changelog

bruvv:
Removing the date fixes the changelog for bashio. REALLY AWESOME addon this btw!
I have a server avaible on oracle cloud that I have avaible if you want to run this (if needed, since a pi is a bit limited in some ways)

### DIFF
--- a/addons_updater/rootfs/scripts/99-run.sh
+++ b/addons_updater/rootfs/scripts/99-run.sh
@@ -138,7 +138,7 @@ for addons in $(bashio::config "addon|keys"); do
     #Update changelog
     touch /data/${BASENAME}/${SLUG}/CHANGELOG.md
     sed -i "1i - Update to latest version from $UPSTREAM" /data/${BASENAME}/${SLUG}/CHANGELOG.md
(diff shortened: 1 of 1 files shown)


alexbelgium:
Hi, thanks for the feedback. What do you mean by fix the changelog? In the past I never had an issue as long as there is an empty line before the date (the addon automatically adds a line above the date when it updates the changelog)

bruvv:
Ah dang it never mind, I thought by naming the version specificly and not adding dates you do not get a huge changelog page. But I was looking at the wrong addon (https://github.com/home-assistant/addons/blob/master/configurator/CHANGELOG.md vs https://github.com/hassio-addons/addon-node-red/releases/tag/v10.3.0 where they use actions to compile a version. NVM closing!

alexbelgium:
No problem! This has run for some time without issues, normally it should be robust. Thanks for being on the lookout. The only addons with a big changelog are the ones for which only nightly builds exist. That's why I run the updates every 2-3 days and not every night as I did in the beginning

PR #190, opened 2022-01-26
Fix cache corruption

woutercoppens:
upgrade-db.php with clear-cache option to fix potential cache corruption. Check-cache doesn't solve all potential problems.


alexbelgium:
Thanks

PR #215, opened 2022-02-13
Update 99-configuration.sh

fkhofmann:
fixed variable definitions/usages for being able using SMB shares

Instead of accessing a mounted SMB share the script created folders for 'storage_dir' and 'import_dir' in the container's filesystem.
The quotation marks of the variable definition where part of the folder names. So it really tried to create a folder like "/mnt/papermerge/storage" (including quotation marks).


alexbelgium:
Thanks very much!

PR #227, opened 2022-02-24
added imap import, improved OCR configuration

fkhofmann:
Papermerge can now be configured to use IMAP for document import.

OCR configuration has been rewritten, first language will be set as Papermerge's default OCR languge


fkhofmann:
ah, let me fix the complains of the linter first.. :)

Issue #241, opened 2022-03-07
[inadyn] forced-update settings is not working


lbouriez:
**Which addon?**

- Addon name : Inadyn
- Addon version : 2.9.1-6

**Describe the bug**
I am trying to setup cloudfare and I used the forced-update parameter, I get this error:

```
/etc/cont-init.d/99-run.sh: executing
/etc/inadyn.conf:4: invalid integer value for option 'forced-update'
Parse error in /etc/inadyn.conf
/etc/cont-init.d/99-run.sh: exiting 1
```

**To Reproduce**

My configuration is:

```
providers:
  - provider: cloudflare.com
    username: XX.org
    password: XX
    hostname: XX.YY.org
forced_update: true
```

alexbelgium:
Hi @x1pher and @deku-m, as I know you're using the addon, are you using forced_update? Thanks!

alexbelgium:
Mmh I think you need to use : 
```yaml
providers:
  - provider: cloudflare.com
    username: XX.org
    password: XX
    hostname: XX.YY.org
    forced_update: true
```

alexbelgium:
Did it work? I see it prevents the error message but not if it actually works :)

lbouriez:
The error is not appearing anymore, I will see if it really updates regularly. If I have any issues, will come back here :) but it should be good!

alexbelgium:
Thanks!

lbouriez:
Forced-update seems to be an integer...
```
-f, --forced-update SEC
             How often the IP should be updated even if it is not changed. The time should be
             given in seconds.
```

alexbelgium:
Thanks for the great troubleshooting! I've merged it

PR #244, opened 2022-03-08
Change the type of forced-update

lbouriez:
From the documentation, forced-update should be an integer representing the time in seconds where Inadyn should update:

https://manpages.ubuntu.com/manpages/bionic/man8/inadyn.8.html

#241


lbouriez:
#241 241

alexbelgium:
Thanks!

PR #309, opened 2022-05-01
Bazarr logo and icon

monkey-debugger:
The logo of sonarr was added to bazarr instead of it's actual logo


alexbelgium:
Thanks very much!

PR #320, opened 2022-05-09
Update README.md

pullaf:
Added another way to integrate the addon with HA that enables toggling/monitoring alternate speeds


alexbelgium:
Thanks very much!

PR #358, opened 2022-06-09
Add Ingress autologin to Calibre-web

met67:
As discussed in the previous days, this PR adds Ingress autologin through the `Reverse Proxy Header Name` feature of Calibre-Web.
Since now Ingress is fully functional, direct exposure of Calibre-web on port 8083 is disabled.
The user used to autologin defaults to `admin` but can be customized.
Also cleaned up some remains from previous tests...

### DIFF
--- a/calibre_web/CHANGELOG.md
+++ b/calibre_web/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Feat : ingress autologin (thanks @met67)
 
 ## 0.6.18-ls164 (09-06-2022)
 - Update to latest version from linuxserver/docker-calibre-web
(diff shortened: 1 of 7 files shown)


alexbelgium:
Amazing, thanks!

met67:
That was fast!!! Did you managed to try it? 😁

/M

Inviato da iPhone

alexbelgium:
I haven't tried it, but read through the commits. 

I think modifying `/defaults/app.db` in the Dockerfile will only work for a first install, so it will probably be needed for existing users to delete their current database and restart the addon to recreate it.

Other than that, indeed I see that you cleaned "useless" code that was remnant from copy/paste or test. For the `upstream backend` for which you changed the port, it is actually not needed as we specify the `proxy_pass` to an ip:port instead of just using the `backend` value.

The only thing I am ambivalent with is the line `allow 172.30.32.2; deny all;` in the ingress.conf. I usually avoid hard-coding the ip, and rather using the `%%interface%%` variable replaced at addon start to ensure that the correct ip is used. It should be indeed 172.30.32.2 but on some configs it jumps by to another ip (I haven't really investigated why). But as long as no one complains, that's good :-)

alexbelgium:
So thanks very much for all that! 

I've just added a comment in the CHANGELOG.md to describe the change for other users. And now that I think about it, we'll need to describe also the new variable in the README.md file

alexbelgium:
Tested, looks good everything is working as intended!

PR #374, opened 2022-06-26
Enabled ingress for tandoor recipes addon

iamkarlson:


alexbelgium:
Hi, it's really not that easy : ingress addition must be supported by a proxy such as nginx or caddy inside the addon that routes traffic, and in a way that works for all functions. As I remember, mealie is not compatible with ingress

iamkarlson:
Yeah, I just tried to debug it with my fork and that's pitty =(

PR #378, opened 2022-07-04
Handle upstream formatting change: spaces not tab

jeffothy:
Upstream lsio commit 7c0eb4e6761f5a48d8ce88facd24e2ed4c16f246 in
github.com/linuxserver/docker-qbittorrent changed the formatting of the
line being patched here and broke the sed replacement. Update the sed
command to match as expected.

The result of this failing sed command was actually a hang on my machine since the command in question would remain `chown ... -R /config` instead of `chown ... -R /config/qBittorrent`.


alexbelgium:
Thanks very much!

PR #413, opened 2022-08-02
Update 20-folders.sh

Deckoz2302:
Current Script changes ownership of /config and /config/addons_config which it should not do.

### DIFF
--- a/jackett/rootfs/etc/cont-init.d/20-folders.sh
+++ b/jackett/rootfs/etc/cont-init.d/20-folders.sh
@@ -1,15 +1,15 @@
 #!/bin/bash
 
-if [ ! -d /share/downloads ]; then
-    echo "Creating /share/downloads"
-    mkdir -p /share/jackett
-    chown -R jackett:jackett /share/jackett
+if [ ! -d /share/Jackett ]; then
+    echo "Creating /share/Jacket"
+    mkdir /share/Jackett
+    chown -R abc:abc /share/Jackett
 fi
 
 if [ -d /config/Jackett ] && [ ! -d /config/addons_config/Jackett ]; then
     echo "Moving to new location /config/addons_config/Jackett"
     mkdir /config/addons_config/Jackett
-    chown -R jackett:jackett /config/addons_config/Jackett
+    chown -R abc:abc /config/addons_config/Jackett
     mv /config/Jackett/* /config/addons_config/Jackett/
     rm -r /config/Jackett
     rm -r /config/jackett
@@ -18,5 +18,5 @@ fi
 if [ ! -d /config/addons_config/Jackett ]; then
     echo "Creating /config/addons_config/Jackett"
     mkdir /config/addons_config/Jackett
-    chown -R jackett:jackett /config/addons_config/Jackett
+    chown -R abc:abc /config/addons_config/Jackett
 fi
--- a/jackett/rootfs/etc/cont-init.d/20-folders.sh
+++ b/jackett/rootfs/etc/cont-init.d/20-folders.sh
@@ -2,14 +2,14 @@
 
 if [ ! -d /share/downloads ]; then
     echo "Creating /share/downloads"
-    mkdir -p /share/downloads
-    chown -R abc:abc /share/downloads
+    mkdir -p /share/jackett
+    chown -R jackett:jackett /share/jackett
 fi
 
 if [ -d /config/Jackett ] && [ ! -d /config/addons_config/Jackett ]; then
     echo "Moving to new location /config/addons_config/Jackett"
-    mkdir -p /config/addons_config/Jackett
-    chown -R abc:abc /config/addons_config/Jackett
+    mkdir /config/addons_config/Jackett
+    chown -R jackett:jackett /config/addons_config/Jackett
     mv /config/Jackett/* /config/addons_config/Jackett/
     rm -r /config/Jackett
     rm -r /config/jackett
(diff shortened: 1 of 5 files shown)


jackett/rootfs/etc/cont-init.d/20-folders.sh
```diff
@@ -1,14 +1,14 @@
...
-    chown -R abc:abc /share/downloads
+if [ ! -d /share/Jackett ]; then
+    echo "Creating /share/Jacket"
+    mkdir /share/Jackett
+    chown -R abc:abc /share/Jackett
 fi
 
 if [ -d /config/Jackett ] && [ ! -d /config/addons_config/Jackett ]; then
     echo "Moving to new location /config/addons_config/Jackett"
-    mkdir -p /config/addons_config/Jackett
```
alexbelgium, line 11: You need the -p, otherwise it won't be able to create /config/addons_config/Jackett if /config/addons_config doesn't exist

alexbelgium:
Hi, thanks very much for contributing to improve the addons !

I had a few questions :
- Why did you remove the parent in mkdir (-p) ? `/config` exists by default, but not `/config/addons_config`.
Therefore, you'll have an error if you install the addon and it executes `mkdir /config/addons_config/sonarr`, that's why you need the `mkdir -p /config/addons_config/sonarr` (same for /share/storage/tv)
- Also, what is the purpose of the `/config/.cache/yarn` ?
- Could you please increment the versions in config.json (usually my nomenclature is that there is the upstream app version, and I increment minor development versions with a "-X" after) ? 

Thanks

jackett/rootfs/etc/cont-init.d/20-folders.sh
Deckoz2302: Everyone of the add-ons like overseer/jackett/radarr/sonarr/transmission-openvpn change the ownership of the parent directories when they start up if you use different pgid and puids for each. For security they shouldn't be running with the same ownership.  With the way the startup scripts are now let's say 

-jackett 399:399
- overseer 400:400
- sonarr 401:401
- radarr 402:402
- transmission-openvpn 403:403

The ownership of the directory is set to the last app that starts up. Effectively making all files available to that pgid:puid which is a security issue. 

This is on supervised hassos on a home assistant blue.

Deckoz2302:
For Overseerr, if you remove the -p effectively making it so it does not change the ownership of /config, then it can no longer create the /config/.cache/yarn folder for Overseerr to run.

jackett/rootfs/etc/cont-init.d/20-folders.sh
alexbelgium: That's strange as the container starts as root. Therefore, the mkdir -p (or mkdir) sets ownership as root:root, and it's the `chown -R abc:abc /config/addons_config/Jackett` command that sets the ownership of the Jackett folder and all subfolder. So there shouldn't be an owner change on /config from this code...

I wonder if the issue is not with remnant code from the upstream container that would change the whole /config folder

alexbelgium: On linuxserver addon, [this line](https://github.com/linuxserver/docker-baseimage-alpine/blob/6ba2997eebdf4f57cc1c3048bcbaad854995b924/root/etc/cont-init.d/10-adduser#L55) (example on alpine) sets the whole config folder. 

I think we should keep the mkdir -p but change that with sed.

For example. `sed -i "s|/config|/config/addons_config/Jackett|g" /etc/cont-init.d/10-adduser` would work in the dockerfile

alexbelgium: Actually, removing the /config altogether would be easier as long as it doesn't impact other addons and folders are created elsewhere. `if [ -f /etc/cont-init.d/10-adduser ]; sed -i "/\/config$/d" /etc/cont-init.d/10-adduser; fi` would work

Deckoz2302: Hi,

Yes that makes sense. I do see that in the dockerfile. Anyway I just was trying to find a way to resolve this across a few of the add-ons you have available. I didn't catch that in the dockerfile. Would you like me to revert the changes, and update the dockerfile and changelog across several of your add-ons and open a new PR, or attach it to this PR?

alexbelgium: Thanks actually I've used a search & replace to change the code in all Dockerfiles of addons based on linuxserver, it's probably the fastest. As it changed many files it could have messed up your PR. Thanks very much for detecting this !

However, we can keep it for the other things you were proposing such as the .cache, correction of overseer and other elements? Also, we can take the opportunity of this PR to increment the versions tags in config.json so that, at least for the addons that interest you, you'll have a new version pushed in the the repo. For the other addons, the change would be implemented when next version would be automatically pushed (usually based on upstream app update).

Would that be ok for you ?

Deckoz2302: Yes that would be fine. Thank you for looking at this with me! 

I can revert most of the changes and keep the /config/.cache/yarn  in and update the version tags. I'll look at this later this evening.

alexbelgium: Thanks! I don't myself use most of the addons so there's things I don't see easily, and I really appreciate your effort in making the addons better

PR #424, opened 2022-08-10
fix(omada): update ports and documentation

ankushg:
- discovery is on *UDP* port 29810, not 
- adds more accurate port descriptions

source: https://www.tp-link.com/us/support/faq/3265/

### DIFF
--- a/omada/config.json
+++ b/omada/config.json
@@ -24,7 +24,7 @@
   ],
   "name": "Omada",
   "ports": {
-    "29810/tcp": 29810,
+    "29810/udp": 29810,
     "29811/tcp": 29811,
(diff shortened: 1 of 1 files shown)


alexbelgium:
Thanks!

alexbelgium:
Hi, I'll look at creating the updater.json when I'll have time, thanks. In the past the upstream image had only a latest tag but no versions, which is not the case anymore

ankushg:
Can confirm that swapping the port from TCP to UDP now has all of my devices being successfully adopted by the HA addon! Thanks! 😎

alexbelgium:
Great, thanks very much!

PR #448, opened 2022-08-29
Deckoz2302 omada host network

Deckoz2302:
Run add_on on host network so outgres is possible. Omada controller does not function with only ingres ports. Omada controller needs to run on the same physical network as devices where the gateway provides cidr spanning.  Ie if run with host_network 'false' outgres is only able to see docker network.

With host_network 'true' outgres relies on the gateway for cidr spanning to discover devices. Ie a subnet with 192.168.1.0/24 the gateway 192.168.1.1 would provide address 192.168.1.1-192.168.1.254. or 192.168.0.0/22 would have the gateway of 192.168.0.1 and provide outgres to 192.168.0.0-192.168.3.254


alexbelgium:
Thanks very much, perfect pull request and we'll documented :)

I've just incremented the addon version to allow the image to build and HA to see the update

Issue #449, opened 2022-08-30
✨ [REQUEST] Gitea


baldarn:
**Which addon?**

[gitea](https://gitea.io/en-us/), self hosted git service

PR #453, opened 2022-08-31
adding gitea

baldarn:
addressing [issue](https://github.com/alexbelgium/hassio-addons/issues/449)


alexbelgium:
Thanks very much for the PR. I've made just a few modifs (added image autobuild, removed armv7)

Have you tried if data persistence is achieved ? In the sense that you don't need additional steps to map the data folder to a remanent place sur as /data, /share or /config ?

gitea/config.json
```diff
@@ -0,0 +1,37 @@
...
+    "keyfile": "privkey.pem",
+    "ssl": false
+  },
+  "ports": {
+    "3000/tcp": 3000,
+    "2222/tcp": 2222
+  },
+  "ports_description": {
+    "3000/tcp": "Http web interface",
+    "22/tcp": "Ssh"
```
alexbelgium, line 26: Hi, there is a mismatch : in port_description your have 22 but in ports you have 2222? thanks!

gitea/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+
+ARG BUILD_FROM=gitea/gitea
+
+FROM ${BUILD_FROM}
+
+##################
+# 2 Modify Image #
+##################
+
+ENV USER git
```
alexbelgium: beware that by using a custom USER to run the addon, it could prevent access to the folders that can be used for data remanence and to install the scripts in section 3. of the Dockerfile ; a way around that is to run the addon as root, but then start the app as your specific user

baldarn: thanks! this is a problem that I was trying to solve now! :)

gitea/config.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "schema": {
+    "certfile": "str",
+    "keyfile": "str",
+    "ssl": "bool"
+  },
+  "slug": "gitea",
+  "url": "https://github.com/baldarn/hassio-addons/tree/master/gitea",
```
alexbelgium: hi, your url is still mapped to your repo and not mine (if you prefer the addon to be added in my repo)

baldarn: yes, I did that to test it on my home assistance instance

gitea/Dockerfile
alexbelgium: I've spent a lot of time also on apps that only accept non-root users to make them work :-) the only way I had found was what I described above - if you find any other way I would be very interested ;-)

gitea/config.json
baldarn: I changed it to 22 => 2222

gitea/Dockerfile
baldarn: Sorry, what are the command to access the folders (i'm trying to access `/share`)?
I removed the `ENV USER git`, but i still can't save things in `/share`

alexbelgium: To access /share, you need to add a map in config.json (I think you did that), but the folder itself doesn't get mapped during addon build as this occurs either on github or on a temporary folder. The usual way is either to change the app data folder through an `ENV` variable (if it exists in gitsea), or to have a script at boot time that creates the relevant folder than symlinks the folders. Here is an example : https://github.com/alexbelgium/hassio-addons/blob/master/scrutiny/rootfs/etc/cont-init.d/01-run.sh#L8:L15

Depending if the addon supports S6 or not, either I add this script in the /etc/cont-init.d folder, or I redirect the container CMD to a specific entrypoint script as in this example : https://github.com/alexbelgium/hassio-addons/tree/master/fireflyiii/rootfs

To make addons updates easier, I use modular codes, in the sense that my scripts (such as the entrypoint) are the same in all addons. Some code is therefore unecessary depending on the addon context, but it allows much easier parallel updates in the future

Usually for my addons, either I use `/data` for data remanence (which allows that data is backuped at the same time as the addon), or `/config/addons_config/ADDONNAME` that is standard for all my addons. `/share` is of course great also if you prefer, as normally `/config` is more for config files than data

baldarn:
I'm pretty satisfied like this for the moment.

if you want anything else done let me know. :)

alexbelgium:
Is data remanence active? Just to know if I can push it to the repo ? Thanks!

gitea/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+ENV GITEA_CUSTOM /share/gitea/data
```
alexbelgium: I'm sorry but actually I think this could create some more issues as gitea would have to create the folder by itself, but not sure the `git` user will have mkdir permission on `/share`. The way I usually solve this is by adding `sed -i "1a mkdir -p /share/gitea/data" /usr/bin/entrypoint` instruction to Dockerhub.

The last solution : I see that the default gitea container location `ENV GITEA_CUSTOM=/data/gitea` would store the data in `/data` which would also allow remanence

baldarn:
> Is data remanence active? Just to know if I can push it to the repo ? Thanks!

I don't know. how can I check that?
the configurations are saved in shared (is this the best place to put them?)
the repos and data in the folder `/data` in the container.

alexbelgium (changes requested):
If data remanence is achieved all is good for me :-)

gitea/Dockerfile
baldarn: perfect, i will remove my custom `ENV GITEA_CUSTOM` and `ENV USER`

alexbelgium:
That's great for me! And this can benefit from this repo tools such as autobuild and autoupdate :-)

alexbelgium:
A visual improvement could be to add transparent png for icons

PR #454, opened 2022-08-31
Corrected version parameter case

acervicius:
lower case version has no effect and produces a warning on startup. upper case produces the correct behavior of skipping the update check and leaving it to docker.


alexbelgium:
Thanks very much!

acervicius:
> Thanks very much!

Thank you for great addon mods!

Issue #474, opened 2022-09-12
🐛 [gitea] Enabling ssl in the addon config does not enable ssl


financelurker:
**Which addon?**

- Addon name : gitea
- Addon version : 1.17.2

**Describe the bug**

If I enable ssl in the addon config the http port just listenes for http traffic.

**To Reproduce**

1. Go to 'addon config'
2. Click on 'ssl'
3. restart the service
4. open the web-ui with https://...
5. See error --> no TLS enabled

**Full addon config**

```yaml
certfile: fullchain.pem
keyfile: privkey.pem
ssl: true
```

**Findings**
To enable ssl in gitea there should be some extension to the `/data/gitea/conf/app.ini` file:
```ini
...
[server]
PROTOCOL  = https
CERT_FILE = /ssl/fullchain.pem
KEY_FILE  = /ssl/privkey.pem
...
```

Tho, the service is run as the "gitea" user, and the "/ssl" mount is only imported, so the "root" user can read the key... This leads to gitea not starting the service (I've tried it by accessing the container and manually adding that config).

See:
```bash
bash-5.1# pwd
/ssl
bash-5.1# ls -lha
total 20K    
drwxr-xr-x    3 root     root        4.0K Aug 13 07:10 .
drwxr-xr-x    1 root     root        4.0K Sep 12 07:09 ..
-rw-r--r--    1 root     root        3.6K Aug 13 07:10 fullchain.pem
-rw-------    1 root     root        1.7K Aug 13 07:10 privkey.pem
```

and

```
bash-5.1# ps aux
PID   USER     TIME  COMMAND
    1 root      0:00 /bin/s6-svscan /etc/s6
   15 root      0:00 s6-supervise gitea
   16 root      0:00 s6-supervise openssh
   17 git       0:06 /usr/local/bin/gitea web
   18 root      0:00 sshd: /usr/sbin/sshd -D -e [listener] 0 of 10-100 startups
   64 root      0:00 bash
   80 root      0:00 ps aux
```

alexbelgium:
Hi, thanks for reporting. Hi @baldarn, any idea ? Thanks!
I'll look at it also a bit later

financelurker:
You're welcome.

And I don't know if this should be a dedicated new issue, but when "ssl" is enabled and you click on "OPEN WEB UI" it's opening the http-url.

Would be nice to configure the base-url in the addon as well (lets say: my dns config is a bit more complex 😁).

[...]

PR #477, opened 2022-09-12
adding more options for gitea (support requested)

baldarn:
hi @alexbelgium ,
I'm trying to solve the https://github.com/alexbelgium/hassio-addons/issues/474 for the ssl.

I was also trying to add some options to better edit the add on from hassio configuration page.

I'm kind of in pain here. I get that you have some best practices, but being a bit noob with docker and pretty young with the project I feel like I would do more damage that good things :'(

Can you give me some advice on how you would address this? in this PR I tried something, but I somehow override the /data folder that the base image of gitea automatically have. this happends adding the automatic scripts that I find in lots of addon, but I don't know where that happens :/

I'm sorry to bother you, but I tried for some hours and got nothing much :(


alexbelgium:
Hi, I really don't have much time for the moment I'll check this weekend! Thanks

alexbelgium:
Hi, you were nearly there. I've looked at the gitea code and changed the last line of the run.sh to execute their entrypoint, then use sed to modify the app.ini. I haven't tested it though yet

alexbelgium:
It's slightly better but still the new values pushed in the app.ini don't seem to be considered... I'll have to look a bit deeper except if you have other ideas?

baldarn:
thanks for the work! I'll try to find some time to test it... maybe tonight... :ocean:

alexbelgium:
> thanks for the work! I'll try to find some time to test it... maybe tonight... 🌊

Honestly I hadn't the time to really test it also, just saw that the code allowed to launch but that the new configs were not taken into account...

alexbelgium:
Hopefully it is fine now

alexbelgium:
Hi, should I merge it? Thanks!

baldarn:
the ssl is working, I would like to add the other vars in the file and then we can merge it ;)

baldarn:
forget about it... I tried editing the confs and the start engine of gitea doesn't take care of it... let's leave those edit manually in the confs for now :/

alexbelgium:
"I tried editing the confs and the start engine of gitea doesn't take care of it..." indeed, even though the values are added in the app.ini they don't seem to be read

alexbelgium:
I've got it : at first runs it takes only variables from /etc/s6/gitea.
I've added them as sed and I'm pretty sure it should work

baldarn:
there is a problem with the version number...meh... tell me if it's ok like this or we have to bump it. consider that I was folliwing the version number of the original gitea project :/

alexbelgium:
> there is a problem with the version number...meh... tell me if it's ok like this or we have to bump it. consider that I was folliwing the version number of the original gitea project :/

For version number you can add what you prefer, there is no specificities. Anyway It will be overwritten at next automatic update. As long as there is a change vs the current number, to allow HA to detect the new version

baldarn:
ok, I changed it, but I see that the check for updated fails. I set ```x.y.z.v``` because `x.y.z` is for the gitea reference and `v` is for the plugin... the tests are not passing :/

alexbelgium:
Don't worry about that, actually I use those tests more as a safeguard than a requirement :) I'll merge it! Let's see how that goes!

alexbelgium:
Does it work for you ? It doesn't load on my side...

baldarn:
is working for me. Maybe you changed the port? I saw that changing the port from 3000 to another doesn't work... :/

alexbelgium:
I've pushed a final version that works, and automatically calculates the root url if it is not encoded manually based on the domain option

Issue #491, opened 2022-09-25
🐛 [Inadyn] Missing dependency mustache-cli


Ealenn:
**Which addon?**

- Addon name : Inadyn
- Addon version : 2.9.1-6

**Describe the bug**

Unable to start this addon due to missing CLI command.

**Full addon log**

```sh
Starting...
/etc/cont-init.d/00-banner.sh: executing
-----------------------------------------------------------
 Add-on: Inadyn
 Inadyn is a small and simple Dynamic DNS, DDNS, client with HTTPS support. A large number of dynamic dns providers are supported (https://github.com/troglobit/inadyn#supported-providers). For those that are not, you can use a custom provider as per this project's documentation (https://github.com/nalipaz/hassio-addons/blob/master/README.md)
-----------------------------------------------------------
 Add-on version: 2.9.1-6
 You are running the latest version of this add-on.
 System: Home Assistant OS 9.0  (aarch64 / raspberrypi4-64)
 Home Assistant Core: 2022.9.6
 Home Assistant Supervisor: 2022.09.1
-----------------------------------------------------------
 Please, share the above information when looking for help
 or support in, e.g., GitHub, forums
 https://github.com/alexbelgium/hassio-addons
-----------------------------------------------------------
/etc/cont-init.d/99-run.sh: executing
/./etc/cont-init.d/99-run.sh: line 7: mustache-cli: command not found
/etc/cont-init.d/99-run.sh: exiting 127
```

**Full addon config**

```yaml
- hostname: mysubdomain.example.com
  provider: domains.google.com
  username: hiddenusername
  password: hiddenpassword
```

**System**

- Raspberry 4
- Home Assistant 2022.9.6
- Supervisor 2022.09.1
- Operating System 9.0
- Interface utilisateur : 20220907.2 - latest

Ealenn:
Dependency used in Dockerfile seem [DEPRECATED](https://github.com/quantumew/mustache-cli)
https://github.com/alexbelgium/hassio-addons/blob/master/inadyn/Dockerfile#L85

Probably not related but 🤷‍♂️

Ealenn:
![image](https://user-images.githubusercontent.com/13718470/192155540-51df3f2b-f6e5-4b89-a262-8cf7e5a6ca25.png)

Ealenn:
Seem related to 

```sh
ealen hassio-addons/inadyn on  master +0 -0 [?] 祥1m33s
[...]

PR #492, opened 2022-09-25
fix(inadyn): go modules management

Ealenn:
Related to https://github.com/alexbelgium/hassio-addons/issues/491

`mustache-cli` isn't installed due to Go 11 modules management.


alexbelgium:
Thanks very much!

PR #497, opened 2022-09-30
Adding local mounting option by uuid

Spand0x:
Added functionality for mounting drives by uuid.
Mounted drives by UUID will never change their ID.


alexbelgium:
Thanks very much! Beware that addons using the cider will either need to be rebuilt locally or on github to incorporate the new code. If there are specific addons you wanted to use without waiting for a version update, we can increment the version to trigger an update

Spand0x:
Hey, Yeah. I am implementing this mainly for Photoprism. So how can we increment the version?

PS:
Thanks for the addons

Issue #502, opened 2022-10-04
🐛 [unpackerr] Unable to change default Folder path 


nathang21:
**Which addon?**

- Addon name : unpackerr
- Addon version : 0.10.1

**Describe the bug**

I am unable to change the default watch folders according to the official unpackerr instructions. My config file is in `/config/unpackerr.conf` and I took the example file from the parent repo, and only changed the `folder.path` and `folder.extract_path` options, which are both set to `/media`. 

My theory is that the addon environmental variables of `UN_FOLDER_0_PATH` and `UN_FOLDER_0_EXTRACT_PATH` are overriding the config file, since the log shows the config file is read. The way the addon is currently setup, there is no way to change the default values without a fork. 

**To Reproduce**

1. Save config file
2. Start addon
3. View log

Expected: Logs to show config file read, and folder /media to be scanned.
Actual: Logs show config file read, but folder still remains set to default Env values of /share/...

**Full addon log**

```
022/10/04 09:56:36.872149 start.go:128: Unpackerr v0.10.0 Starting! (PID: 7) 2022-10-04 09:56:36.869420559 -0400 EDT m=+0.002993323
2022/10/04 09:56:36.872312 logs.go:210: ==> GoLift Discord: https://golift.io/discord <==
2022/10/04 09:56:36.872333 logs.go:211: ==> Startup Settings <==
2022/10/04 09:56:36.872365 sonarr.go:87:  => Sonarr Config: 0 servers
2022/10/04 09:56:36.872382 radarr.go:88:  => Radarr Config: 0 servers
2022/10/04 09:56:36.872397 lidarr.go:87:  => Lidarr Config: 0 servers
2022/10/04 09:56:36.872412 readarr.go:87:  => Readarr Config: 0 servers
2022/10/04 09:56:36.872485 folder.go:69:  => Folder Config: 1 path: /share/downloads_packed, extract to: /share/downloads_unpacked (delete after:0s, delete orig:false, log file: true, move back:false, event buffer:20000)
2022/10/04 09:56:36.872511 logs.go:217:  => Using Config File: /config/unpackerr.conf
2022/10/04 09:56:36.872527 logs.go:218:  => Parallel: 1
2022/10/04 09:56:36.872541 logs.go:219:  => RAR Passwords: 0
2022/10/04 09:56:36.872558 logs.go:220:  => Interval: 2m
2022/10/04 09:56:36.872591 logs.go:221:  => Start Delay: 1m
[...]

PR #504, opened 2022-10-04
Remove comment in example to enable the folder configuration itself.

nathang21:
Remove comment in example to enable the folder configuration itself.

Related: #502

### DIFF
--- a/unpackerr/README.md
+++ b/unpackerr/README.md
@@ -51,7 +51,7 @@ In /config/unpackerr.conf you can set all variables according to this list of en
 
 Folders must be custimozed in the conf file with the lines :
 ```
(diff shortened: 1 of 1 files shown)


alexbelgium:
Thanks! Does this work like that?

nathang21:
According to [this comment](https://github.com/davidnewhall/unpackerr/blob/master/examples/unpackerr.conf.example#L58) it seems to suggest so.

If you mean does this resolve #502, not yet. It tries to start folder watching, but still fails to find the configured folder.

alexbelgium:
Thanks, I'll try too. I thought that removing the hard codes variable would allow the app to read the config file

PR #532, opened 2022-10-29
Aligned config with upstream

bruvv:
added homeassistant example

### DIFF
--- a/epicgamesfree/CHANGELOG.md
+++ b/epicgamesfree/CHANGELOG.md
@@ -1,11 +1,17 @@
+## 30-10-2022
+
+- New config.js update
 
 ## ee3dff4b29757e28282cada9bcb3e8f84f1c3041-2022-10-23 (2022-10-23)
+
 - Update to latest version from charlocharlie/epicgames-freegames
 
 ## 80fdb2447c8846bee011c0d6d5c44c56210cb24a-2022-09-23 (2022-09-23)
+
 - Update to latest version from charlocharlie/epicgames-freegames
 
 ## 9531608362dc8c1232c1e12005541796610d2af7-2021-12-17 (2021-12-17)
+
 - Update to latest version from charlocharlie/epicgames-freegames
 
 ## 4.1.0 (21-07-2022)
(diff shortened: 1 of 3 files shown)


alexbelgium:
Hi, thanks, it seems to be an invalid json though as ’//’ comments are not accepted I think in json ? Or does the app allows this ?

bruvv:
It is a config example that people should adjust anyway otherwise the app will not work. I will see what I can do to change it to a gui and fix the linting.

bruvv:
@alexbelgium the linting is not something I seem to understand. Can you have a look ?

alexbelgium:
I've learned by myself so it could be incorrect, but linting is to see if the code is aligned with standard. It checks automatically every code update to see if it conforms, is there is a red cross. For json, here is an external linter to check validity : https://jsonlint.com/

Clicking on the red cross next to your commits, you should see the linting performed and the pointers to solve it

Hope this helps!

alexbelgium:
Actually having checked the GitHub action log I think it's more the appearance of passwords on the json (even fake) that trigger a GitLeak alarm

bruvv:
@alexbelgium correct: https://github.com/alexbelgium/hassio-addons/actions/runs/3355107407/jobs/5559129204 this is because of false positive so we can ignore this.
But this: https://github.com/alexbelgium/hassio-addons/actions/runs/3355107404/jobs/5559129826 I do not understand, it is saying probably that I am using tabs instead of spaces which is not true since I am always using 2 spaces instead of 1 tab.

alexbelgium:
> @alexbelgium correct: https://github.com/alexbelgium/hassio-addons/actions/runs/3355107407/jobs/5559129204 this is because of false positive so we can ignore this. But this: https://github.com/alexbelgium/hassio-addons/actions/runs/3355107404/jobs/5559129826 I do not understand, it is saying probably that I am using tabs instead of spaces which is not true since I am always using 2 spaces instead of 1 tab.

It's strange, I do see one tab and not two spaces. I've changed it in the config.json above for example

To ignore gitleaks I'll check how to do that

alexbelgium:
I'll merge it, we'll manage gitleaks afterwards ;) thanks!

PR #546, opened 2022-11-11
Update config.json

novisys:


alexbelgium:
Thanks very much. There is an issues too with custom configs I'm trying to solve. Have a nice day

PR #572, opened 2022-11-23
Add scrutiny host ID details

### LINKED ISSUE #571: ✨ [REQUEST] Scrutiny (FA) - Add Host to Collector Information
**Which addon?**

Scrutiny (And Scrutiny_fa!)

**Is your feature request related to a problem? Please describe.**

At the moment, the collector data is sent without a hostname, it would be good to include this in situations where multiple hosts are sending data to the same Scrutiny "hub"

\*\*If a new addon, have you checked on Google that such as addon doesn't already exists?

**Describe the solution you'd like**

Add the environment variable to enable the hostname to be sent

**Describe alternatives you've considered**

N/A - there's only one way to do this really!

**Additional context**

Current dashboard:

image

proffalken:
Fixes #571 

When running Scrutiny in a hub/spoke model (multiple collectors feeding data back to a single "controller" that runs the UI), the hostname is not currently sent with the data.

This PR aims to fix that by creating [a new environment variable](https://github.com/AnalogJ/scrutiny/blob/master/docs/TROUBLESHOOTING_DEVICE_COLLECTOR.md#hub--spoke-model-with-multiple-hosts) that matches the value in the Scrutiny documentation, and sets it by default to home_assistant.

NOTE: This is currently **UNTESTED** as I cannot get a local dev environment working.

### DIFF
--- a/scrutiny/config.json
+++ b/scrutiny/config.json
@@ -44,7 +44,8 @@
     "/dev/nvme2"
   ],
   "environment": {
-    "COLLECTOR_API_ENDPOINT": "http://localhost:8080"
+    "COLLECTOR_API_ENDPOINT": "http://localhost:8080",
+    "COLLECTOR_HOST_ID": "home_assistant"
   },
   "ingress": true,
   "init": false,
@@ -71,6 +72,7 @@
   ],
   "schema": {
     "COLLECTOR_API_ENDPOINT": "str?",
+    "COLLECTOR_HOST_ID": "str?",
     "expose_config": "bool?",
     "Mode": "list(Collector+WebUI|Collector)?",
     "TZ": "str?",
--- a/scrutiny_fa/config.json
+++ b/scrutiny_fa/config.json
@@ -7,7 +7,8 @@
   "codenotary": "alexandrep.github@gmail.com",
   "description": "Scrutiny WebUI for smartd S.M.A.R.T monitoring (Full Access)",
   "environment": {
-    "COLLECTOR_API_ENDPOINT": "http://localhost:8080"
+    "COLLECTOR_API_ENDPOINT": "http://localhost:8080",
+    "COLLECTOR_HOST_ID": "home_assistant"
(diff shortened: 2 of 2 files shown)


alexbelgium:
Looks great! Thanks. We could even set the hostname through an addon option if you'd like. It would require to pass the ENV variable as an "export variable" directly at the start of the script

proffalken:
> Looks great! Thanks. We could even set the hostname through an addon option if you'd like. It would require to pass the ENV variable as an "export variable" directly at the start of the script

Ah, I thought I'd done that in this PR - looks like I've got a lot more to learn about how Addons work! :rofl: 

I'll have another look and see if I can work out how to do that as well.

Thanks for merging it so quickly!

alexbelgium:
Hi, I read it too quickly !

The "environment" tag in the config.json file is the same as an ENV command in a Dockerfile : it sets it at a container level.
The "COLLECTOR_HOST_ID": "str?" in the "schema" tag makes it appear as optional in the addons options. Once you set it, it is set in the /data/options.json file. Then, the way to access it is with bashio commands. Bashio is a small app made by one of the HA developer to automatize elements. For example, `bashio::config.has_value "COLLECTOR_HOST_ID"` would check if `COLLECTOR_HOST_ID` is set.

The right commands could be something like : 
    if bashio::config.has_value "COLLECTOR_HOST_ID"; then
        echo "export COLLECTOR_HOST_ID=$(bashio::config "COLLECTOR_HOST_ID")" >> /env.sh
        sed -i "1a export COLLECTOR_HOST_ID=$(bashio::config "COLLECTOR_HOST_ID")" /etc/services.d/collector-once/run
        bashio::log.info "Using 'COLLECTOR_HOST_ID' $(bashio::config "COLLECTOR_HOST_ID")"
    fi

proffalken:
Ah, cool, I'll try and get that added in the next few days. Thanks!

PR #594, opened 2022-12-09
Mailer security changed

NODeeJay:
Hi Alex,

with version 2.7 the variable was changed from MAILER_SECURE to MAILER_SECURITY and now it accepts a string with either tls, starttls, or none.

I hope I found it everywhere :-)

### DIFF
--- a/joplin/CHANGELOG.md
+++ b/joplin/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Breaking change: MAILER_SECURE replaced by MAILER_SECURITY
 - WARNING : update to supervisor 2022.11 before installing
 - Add access to systime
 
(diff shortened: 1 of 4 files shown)


alexbelgium:
Thanks! You'll just also need to increment the version in config.json `"version": "2.9.7-3",` to allow the new version to be shown in HA

NODeeJay:
Will it be then a 2.9.8 (following Semantic versioning) or 2.9.7-4?

alexbelgium:
> 

In my naming convention, I align with the upstream app. It is still at 2.9.7, so I usually increment by adding something after this number. Thus my 2.9.7-3 that doesn't follow semantic versioning, but that will be replaced by 2.9.8 as soon as it is emitted by the upstream maintainer

alexbelgium:
Thanks! As I saw in the readme that there were only 3 possible values for the new parameter, I've made it a list instead of a str. Don't hesitate to change back if this is not correct

NODeeJay:
List is a good idea, thanks, did not consider that. Anything left or shall we release it?

alexbelgium:
Perfect, thanks!

PR #611, opened 2022-12-22
fix docker version gitea

baldarn:
my bad. I tried updating but I saw that I went on the 1.18.0 rc1. :/


alexbelgium:
Hi, you'll need to publish a new version too in the config.json for the Addon to build (like 1.17.4-2) thread's

alexbelgium:
Thanks

PR #612, opened 2022-12-23
Fix config template for Plex meta manager

bruvv:
I changed the url to the propper url, now it will download the file instead of the HTML.


alexbelgium:
Thanks!

Issue #674, opened 2023-01-27
[Paperless] config.yaml application


alexbelgium:
I'm running into trouble having the Paperless addon use a network mount for data storage. Settings are as follows:

In the addon's config UI:
`networkdisks`: `//192.168.1.5/Paperless`
`CONFIG_LOCATION`: `/config/addons_config/paperless_ng/config.yaml`

In `/config/addons_config/paperless_ng/config.yaml`:
```yaml
PAPERLESS_CONSUMPTION_DIR: /mnt/Paperless/consume
PAPERLESS_DATA_DIR: /mnt/Paperless
PAPERLESS_MEDIA_ROOT: /mnt/Paperless/media
```

It looks like the the config.yaml entries are overridden by [these lines](https://github.com/alexbelgium/hassio-addons/blob/3121a0e8248f4914c3d3ecc3a359aedf2e24ef97/paperless_ngx/Dockerfile#L43-L46) in the Dockerfile. Running `printenv` from a shell in the addon's container returns the locations from the Dockerfile, not from my config file.

Are these folders deliberately hard coded? Is there a way to get the Paperless addon to use network mounts for consumption, data, and media?

Thanks!

_Originally posted by @elahd in https://github.com/alexbelgium/hassio-addons/discussions/673_

alexbelgium:
Could you please add your addon logs? Thanks

alexbelgium:
Actually I noticed I have the same setting and it works for me

elahd:
Thanks for looking into this. More details below. I redacted some sensitive information with 'XXX's.

Here are the logs:

```
Starting...
/etc/cont-init.d/00-banner.sh: executing
-----------------------------------------------------------
 Add-on: Paperless NGX
 scan, index and archive all your physical documents
-----------------------------------------------------------
 Add-on version: 1.11.3-2
 You are running the latest version of this add-on.
 System: Home Assistant OS 9.4  (aarch64 / raspberrypi4-64)
 Home Assistant Core: 2023.1.7
 Home Assistant Supervisor: 2023.01.1
-----------------------------------------------------------
 Please, share the above information when looking for help
 or support in, e.g., GitHub, forums
 https://github.com/alexbelgium/hassio-addons
-----------------------------------------------------------
/etc/cont-init.d/01-custom_script.sh: executing
[16:38:56] INFO: Execute /config/addons_autoscripts/paperless-ng.sh if existing
[16:38:56] INFO: ... no script found
[...]

PR #685, opened 2023-02-03
Added paperless folder config options to addon config.

elahd:
This is a proposal to address #674.


alexbelgium:
Hi, thanks for the proposal, we'll just have to look to also add code to create the folders and provide appropriare permissions (if the app doesn't do it by itself)

alexbelgium:
Thanks!

PR #701, opened 2023-02-08
adding cryptpad

baldarn:
please, take a look.

I didn't test if any of this is working. I don't have time now.

so, If you want to try it and test it cool! if not, wait till I try it and everything is ok ;)


alexbelgium:
Thanks! I don't have much time but the code looks great :-) If you want to fork and test it, you'll need to remove the "image" tag from the config.json which will allow you to locally build it on your HA. And we can re-add the "image" tag once it's complete

alexbelgium:
Btw (always on the idea of avoiding duplication) it seems to me it's an office app based on crypto. Have you evaluated using nextcloud using the OTFE encryption module for this aim ?

alexbelgium:
You can also use the "Lint on change" checks to help with debugging, that provide infos such as :
`2023-02-08 14:18:29 [INFO]   File:[/github/workspace/cryptpad/rootfs/etc/cont-init.d/99-run.sh]
2023-02-08 14:18:29 [ERROR]   Found errors in [shellcheck] linter!
2023-02-08 14:18:29 [ERROR]   Error code: 1. Command output:
------

In /github/workspace/cryptpad/rootfs/etc/cont-init.d/[99](https://github.com/alexbelgium/hassio-addons/actions/runs/4125028404/jobs/7125043643#step:4:100)-run.sh line 16:
CPAD_TLS_CERT="/ssl/$(bashio::config 'certfile')" "$file"
                                                   ^---^ SC2154 (warning): file is referenced but not assigned (for output from commands, use "$(file ...)" ).`

baldarn:
> Btw (always on the idea of avoiding duplication) it seems to me it's an office app based on crypto. Have you evaluated using nextcloud using the OTFE encryption module for this aim ?

I'm really not sure.
nextcloud is mastodontic compared with this... I don't know. if you think this is duplicated I'll just drop the pr.
your call ;)

alexbelgium:
Nono, just checking ;) indeed it does seem a quite differznt purpose and I also avoided nextcloud apps to use specific ones that are usually more optimized

alexbelgium:
This image only supports amd64, do you now if there is a nice one that would support both amd64 and aarch64

baldarn:
I'm not really practical in different architectures and how to fix this. maybe for the first integration we can keep the amd64 and then see if someone needing it for the aarch64 arrives ;)

baldarn:
still not working... I miss the docker knowledge to run the main application. It seems the configs are now correctly updated, but it freezes wating for the main app... :/

when I will have more time I'll try to fix it ;)

alexbelgium:
That's normal : you are using my entrypoint logic which only executes scripts in /etc/cont-init.d
Therefore, once the *run.sh is executed, it stops. If using the entrypoint logic, your last line in the *run.sh file should be the execution of the app. You can find the appropriate commabd by looking at thz ENTRYPOINT or CMD from the upstream container

alexbelgium:
It is https://github.com/xwiki-labs/cryptpad-docker/blob/02734eba28fc946f54fbb5ca4a9840ecb3ec491d/Dockerfile#L51 do you should have : ```
cd /cryptpad || true
npm start
```

alexbelgium:
BTW you can use the Checks tab to have info on elements that non optimal in the scripts

PR #727, opened 2023-02-22
[Calibre] Add wireless connection port to config.json

danbruno:
Tried my best to follow the guidelines given. I've tried this on my local home assistant instance and it works as expected. My Kobo Libra 2 can connect wirelessly to calibre using Kobo-UNCaGED which is really handy when my add-on server is tucked away and the USB ports aren't easily accessible.

The guidelines weren't too descriptive on how to increase the version number so I apologize if I got that incorrect.

### DIFF
--- a/calibre/CHANGELOG.md
+++ b/calibre/CHANGELOG.md
@@ -1,3 +1,5 @@
+## v6.13.0-ls204 (21-02-2023)
+- Add wireless connection port
 
 ## v6.13.0-ls203 (19-02-2023)
 - Update to latest version from linuxserver/docker-calibre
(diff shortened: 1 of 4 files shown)


alexbelgium:
Thanks very much! On terms of versioning, i align with the upstream image (base container for my addon). So as "v6.13.0-ls203" comes from the upstream image, I usually increment with -x to show my own subversions in between upstream updates.

danbruno:
> Thanks very much! In terms of versioning, i align with the upstream image (base container for my addon). So as "v6.13.0-ls203" comes from the upstream image, I usually increment with -x to show my own subversions in between upstream updates.

Awesome, thanks so much!

PR #779, opened 2023-04-03
Use utf-8 as default charset encoding in CIFS mounts

leokeba:
Hi, I'm having issues with accents in filenames using the cifs mounting options in your addons, this is mostly annoying with qBittorrent as I have a lot of music torrents with lots of files and lots of accents and I can't seed them correctly or download new files with proper names.

Please do not merge this without proper review / testing, I think using the `iocharset=utf8` option in the mount command might be all that's needed to fix the issue, but I wasn't able to test it myself as I am proficient enough in docker stuff to properly setup my own functional repo.

I tried messing directly with the containers on my install but haven't had any success so far, do you have a suggestion on how I could easily test it ?

### DIFF
--- a/.templates/92-smb_mounts.sh
+++ b/.templates/92-smb_mounts.sh
@@ -67,17 +67,28 @@ if bashio::config.has_value 'networkdisks'; then
 
         # if Fail test different smb and sec versions
         if [ "$MOUNTED" = false ]; then
-            for SMBVERS in ",vers=3" ",vers=1.0" ",vers=2.1" ",vers=3.0" ",nodfs" ",uid=0,gid=0,forceuid,forcegid" ",noforceuid,noforcegid" ",domain=${DOMAIN:-WORKGROUP}" ",noserverino"; do
-                mount -t cifs -o "iocharset=utf8,rw,file_mode=0775,dir_mode=0775,username=$CIFS_USERNAME,password=${CIFS_PASSWORD}$SMBVERS$PUID$PGID" "$disk" /mnt/"$diskname" 2>/dev/null && MOUNTED=true && break || MOUNTED=false
-                for SECVERS in ",sec=ntlmv2" ",sec=ntlmv2i" ",sec=ntlmssp" ",sec=ntlmsspi" ",sec=krb5i" ",sec=krb5" ",iocharset=utf8" ",noserverino"; do
-                    mount -t cifs -o "iocharset=utf8,rw,file_mode=0775,dir_mode=0775,username=$CIFS_USERNAME,password=${CIFS_PASSWORD}$SMBVERS$SECVERS$PUID$PGID" "$disk" /mnt/"$diskname" 2>/dev/null && MOUNTED=true && break 2 && break || MOUNTED=false
+        
+            # Test with PUIDPGID, remove otherwise
+            ######################################
+            for PUIDPGID in "$PUID$PGID" ""; do
+            
+                # Test with iocharset utf8, remove otherwise
+                ############################################
+                for CHARSET in ",iocharset=utf8" ""; do
+                
+                    # Test with different SMB versions
+                    ##################################
+                    for SMBVERS in ",vers=3" ",vers=1.0" ",vers=2.1" ",vers=3.0" ",nodfs" ",uid=0,gid=0,forceuid,forcegid" ",noforceuid,noforcegid" ",domain=${DOMAIN:-WORKGROUP}"; do
+                        mount -t cifs -o "rw,file_mode=0775,dir_mode=0775,username=$CIFS_USERNAME,password=${CIFS_PASSWORD}$CHARSET$SMBVERS$PUIDPGID" "$disk" /mnt/"$diskname" 2>/dev/null && MOUNTED=true && break || MOUNTED=false
+                        
+                        # Test with different security versions
+                        #######################################
+                        for SECVERS in ",sec=ntlmv2" ",sec=ntlmv2i" ",sec=ntlmssp" ",sec=ntlmsspi" ",sec=krb5i" ",sec=krb5"; do
+                            mount -t cifs -o "rw,file_mode=0775,dir_mode=0775,username=$CIFS_USERNAME,password=${CIFS_PASSWORD}$CHARSET$SMBVERS$SECVERS$PUIDPGID$CHARSET" "$disk" /mnt/"$diskname" 2>/dev/null && MOUNTED=true && break 2 && break || MOUNTED=false
+                        done
+                    done
                 done
             done
-        fi
-
-        # if Fail test without PUID and PGID
-        if [ "$MOUNTED" = false ]; then
-            mount -t cifs -o iocharset=utf8,rw,username="$CIFS_USERNAME",password="${CIFS_PASSWORD}" "$disk" /mnt/"$diskname" && MOUNTED=true || MOUNTED=false
         fi
 
         # Messages
--- a/.templates/92-smb_mounts.sh
+++ b/.templates/92-smb_mounts.sh
@@ -63,21 +63,21 @@ if bashio::config.has_value 'networkdisks'; then
 
         # Tries to mount with default options
         # shellcheck disable=SC2140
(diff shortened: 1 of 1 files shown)


alexbelgium:
Hi, thanks very much for the PR. Actually I've implemented a way to test code modifications without changing a full version (ok I must confess sometimes I push a new version just to test some code...)

You need to :
- Create a file named /config/addons_autoscripts/qbittorrent.sh
- Include the following code inside : 
```bash
#!/bin/bash
echo "Starting script"
sed -i "s|cifs -o \"rw,|cifs -o \"rw,iocharset=utf8,|g" /etc/cont-init.d/92-smb_mounts.sh
echo "... done"
```
- restart the addon

alexbelgium:
What could break if we implement iocharset by default? Is there any potential issue ?

leokeba:
Thanks, that's very useful.

I've tried that way and it breaks the script, dmesg output gives the following error :
`CIFS: VFS: CIFS mount error: iocharset utf8 not found`
That seems similar to the issue raised here : [](https://askubuntu.com/questions/519796/unable-to-mount-cifs-with-iocharset-utf8)
I tried a few things but I wasn't able to fix it, I guess since it's a kernel module missing a script won't be enought and we might need to modify the docker image, am I right ?

leokeba:
I also found this discussion that may be relevant : https://askubuntu.com/questions/760058/what-is-the-purpose-of-iocharset-utf8

But since I am not able to test it, I don't even know if this is the real issue I'm facing in the first place.

Also, I'm guessing that if it does fix my problem, then it might break previously created paths and filenames. So even if this ends up being a better default option, you might want to be careful with backwards-compatibility.

alexbelgium:
Strange. I have executed the exact same script as above, and have this result :
```
cont-init: info: running /etc/cont-init.d/92-smb_mounts.sh
Mounting smb share(s)...
Using PUID 0 and PGID 0
[09:12:17] INFO: ... //192.168.178.23/NAS/downloads successfully mounted to /mnt/downloads with options 
cont-init: info: /etc/cont-init.d/92-smb_mounts.sh exited 0
```

Therefore, mounting with iocharest=utf8 works. It also works when I log directly in qbittorrent using portainer and mount it manually. I can't explain why it wouldn't work on your system...

alexbelgium:
Are you on HA os ? Or isn't the missing kernel part from your own system ?

leokeba:
Hmm, first I forgot to tell you that your little script had an error in it, it didn't replace some of the commands in the file, here's the command I ended up using :
`sed -i "s|cifs -o |cifs -o iocharset=utf8,|g" /etc/cont-init.d/92-smb_mounts.sh`

I am on HA OS, running in a dedicated VM, fresh install from last week. Can you show me how you did it manually ?

alexbelgium:
Hi, thanks for the feedback. Here is what I did :
- Install portainer addon from my repo
- Disable secure mode in the addon options, then start the addon
- Go in the containers list, then in the qbittorrent container
- Enter the console mode from this screen
- Type your commandes. Here is the one I tried :
```bash
# Adapt variables
CIFS_DISK="//192.168.178.23/NAS"
CIFS_USERNAME="homeassistant"
CIFS_PASSWORD="Bonjour01"

# Mount
mkdir -p /a
mount -t cifs -o "rw,username=$CIFS_USERNAME,password=$CIFS_PASSWORD,iocharset=utf8" "$CIFS_DISK" /a
touch /a/é
ls /a

# Clean
rm /a/é
umount /a
```

alexbelgium:
Btw I've made a proposal of a slightly modified code on your PR to first test with iocharset=utf8 then without if it doesn't work... Your opinion is welcome if you think this is a good way to do, this would allow to still mount even if iocharset is not supported in some addons

alexbelgium:
Regarding the system, I'm using aarch64 on an rpi4. I'll try also on an amd64 VM system to check, but I don't see why it should differ...

I'm not a super expert on docker systems, but I think that the mount occurs at container level, and therefore the container's kernel should be the one used and not the system's one.

Regarding breaking paths... Well I'm more for long term addons quality, and I'll make a disclaimer on the changelogs to say to be aware of it... I do think that your proposal of including this parameter leads to a greater overall quality

leokeba:
Hmm, what version of haos are you running ? I am on 9.5

I was checking the latest releases and just saw that this [PR](https://github.com/home-assistant/operating-system/pull/2446) was just merged, maybe it will solve my issue ? Do you know if there is an easy way to upgrade haos or do I have to backup and reinstall everything just to try it ?

alexbelgium:
Just thought about it : actually the mount messages are in the host's log ; so indeed it is at host and not container level. I'll try on my VM system this evening to see if I can replicate

alexbelgium:
> Hmm, what version of haos are you running ? I am on 9.5
> 
> I was checking the latest releases and just saw that this [PR](https://github.com/home-assistant/operating-system/pull/2446) was just merged, maybe it will solve my issue ? Do you know if there is an easy way to upgrade haos or do I have to backup and reinstall everything just to try it ?
> 
> BTW after reading what I could grab on the subject, it seems that docker images use the kernel of the host, so it probably means something is wrong in haos itself.

Thanks, regarding updating you should receive all updates requests directly in your HA instance. It's OTA updates even in a VM. I'm also on HAos 9.5 though, so I'm not sure what could happen...

leokeba:
Yeah I just had to join the beta channel to get the latest updates. So I installed haos v10, and tried to connect to the cifs share directly from the host, the mount command is indeed present, but I wasn't able to make it work, I get `permission denied` each time I run the command, even without the iocharset option.

Looks like I may be a bit out of my league, which would be really sad as I was really close to setting up the perfect home server for managing all my digital needs, but it looks like I will have to wait for that. I may create an issue in the haos repo if I have time to investigate further, but I already spent waaaay to much time on this so we'll see about that.

Thanks a lof for your help so far, of course if you have other suggestion don't hesitate, and otherwise we can still merge this, but I won't be able to properly test it or even "enjoy" it.

leokeba:
It's running on a Freebox Delta Server, or a kind of router on steroids that my ISP provides, it seems to run qemu, which I saw listed here : https://floss.freebox.fr/freebox_server_delta/4.3.0/index.html

It's not very powerful cpu-wise but a quite capable machine, it has a 10G connexion to the internet and a nas bay with 2 hdd and 2 ssd in raid0 pairs, and with enough swap it runs thousands of torrents and a bunch of docker containers with no problem.

There's not much I can configure myself, except a cloud-init file but I don't think the haos image can make use of it. It's a good lead, but I have another VM running a bunch of stuff that I am currently in the process of migrating to haos, and this one has no issues with cifs mounts or charset encoding, using the same server of course, so I think it proves the host is not the problem. Actually, the other VM makes use of the cloud-init to configure the cifs mounts.

leokeba:
BTW, the latest commit on the haos repo : https://github.com/home-assistant/operating-system/commit/19e2869261b370295c2916188f4c80e510536e2d

leokeba:
I kept investigating and found the issue is clearly specific to the haos aarch64, I tried everything we suggested again from scratch on different machines using different releases and the issue comes up on both aarch64 platforms I tried, but it does not happen on my intel mac.

As it turns out, the cifs mount command works properly even without specifying utf8 charset so it must be using it by default anyway on x86/64 archs. What that says to me is that it definitely caused by a kernel module or something similar from the host buildroot distribution, so I don't think there is anything you can do, I just have to check if the next release fixes it or raise the issu on their repo.

Thanks again for the help, I guess all of this makes this PR irrelevant, or does it ?

alexbelgium:
Well, this work did lead to uncovering an issue with the aarch64 build and helped improve the code to either specify utf8 charset or not so I would say this is very useful! I still propose to merge the change, seems like a useful safety feature to have utf8 specified even if it is a default for some distributions. Thanks very much for all this work you did, this will both lead to an I'm improvement of the global code and will help other people with the same configuration

leokeba:
Okay, is there something I need to do ? Not sure how to resolve the conflict now, maybe it's time to learn a bit more about git too :)

alexbelgium:
I reverted my change ;-) indeed you are right I should have waited for your merge first. Thanks very much for your contribution!

PR #820, opened 2023-04-30
add renderD128 to devices

jpgimenez:
allows to use hardware acceleration for transcoding on intel processors
https://github.com/pulsejet/memories/wiki/HW-Transcoding#va-api


alexbelgium:
Thanks very much

Issue #749, opened 2023-03-06
🐛 [Calibre] Calibre don't start with samba share mount configured


marlonqpa:
**Which addon?**

- Addon name : Calibre
- Addon version : v6.13.0-ls204-2

**Describe the bug**

Hi 
After last update addon refuse to start. Last line in log informs about correct samba mounting. When I remove mounting option from configuration, addon is starting properly.

**To Reproduce**

**Full addon log**

```
cont-init: info: running /etc/cont-init.d/00-banner.sh
[migrations] started
[migrations] no migrations found
-----------------------------------------------------------
 Add-on: Calibre
-------------------------------------
          _         ()
         | |  ___   _    __
         | | / __| | |  /  \
         | | \__ \ | | | () |
         |_| |___/ |_|  \__/
Brought to you by linuxserver.io
-------------------------------------
To support LSIO projects visit:
https://www.linuxserver.io/donate/
-------------------------------------
GID/UID
-------------------------------------
User uid:    0
User gid:    0
-------------------------------------
 e-book manager
-----------------------------------------------------------
 Add-on version: v6.13.0-ls204-2
 You are running the latest version of this add-on.
 System: Home Assistant OS 9.5  (aarch64 / odroid-c4)
 Home Assistant Core: 2023.3.1
 Home Assistant Supervisor: 2023.03.0
-----------------------------------------------------------
 Please, share the above information when looking for help
 or support in, e.g., GitHub, forums
 https://github.com/alexbelgium/hassio-addons
-----------------------------------------------------------
cont-init: info: /etc/cont-init.d/00-banner.sh exited 0
cont-init: info: running /etc/cont-init.d/00-data_location.sh
[09:29:14] INFO: Setting data location to /config/addons_config/calibre
sed: couldn't edit /etc/s6-overlay/s6-rc.d/99-ci-service-check/dependencies.d: not a regular file
sed: can't read /etc/services.d/web/run: No such file or directory
sed: can't read /etc/services.d/web/run: No such file or directory
usermod: no changes
Creating /config/addons_config/calibre
[09:29:14] INFO: Setting ownership to 0:0
cont-init: info: /etc/cont-init.d/00-data_location.sh exited 0
cont-init: info: running /etc/cont-init.d/01-custom_script.sh
[09:29:14] INFO: Execute /config/addons_autoscripts/calibre.sh if existing
[09:29:14] INFO: ... no script found
[...]

PR #826, opened 2023-05-06
Fix sed error in calibre-web

mechanarchy:
Same fix as #749, commit cae22cc


alexbelgium:
Thanks, and I've incremented the version number to start a new addon version build

PR #872, opened 2023-06-14
Update HA example code for scrutiny readmes

nkm8:
Using the original examples, you only get 1 sensor in Home Assistant that has a hard-coded value of `OK` with all of the JSON data listed as attributes on that single sensor. I split these out to be individual sensors in HA that each have the correct value(s). I also attempted to annotate the special sensors like SMART status and temperature with the correct properties in HA.


alexbelgium:
You're the best thanks!

PR #886, opened 2023-06-28
Create DOCS.md for Tandoor Recipes

nough:
gives more of an explanation of why things are the way they are in the Tandoor Recipes Addon.


alexbelgium:
Thanks very much!

PR #899, opened 2023-07-13
Remove webui

iBobik:
Postgres has no web ui, so do not offer button to it.

### DIFF
--- a/postgres/CHANGELOG.md
+++ b/postgres/CHANGELOG.md
@@ -1,1 +1,2 @@
 - Initial release
+- Removed useless webui button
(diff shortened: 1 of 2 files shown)


iBobik:
I am not sure if my change is related to the fail.

alexbelgium:
Looks like the fail is due to the build script so not relevant here. Thanks very much for the PR!

alexbelgium:
I wanted to implement an admin app such as padmin but it didn't seem easy to do...

iBobik:
I also like the idea of having webui, but I think it is not worth the maintenance and additional memory - most users will need to just set one password and default database for that one app they need it. Others can use one of the many GUI clients remotely connected.

PR #905, opened 2023-07-17
[Draft] remove password from addon updater

DDanii:
Github is no longer allow pushing commits with password.
It's a draft as I didn't tested it yet.

### DIFF
--- a/addons_updater/README.md
+++ b/addons_updater/README.md
@@ -68,10 +68,9 @@ Here you define the values that will allow the addon to connect to your reposito
 ```yaml
 repository: 'name/repo' coming from github
 gituser: your github username
-gitpass: your github password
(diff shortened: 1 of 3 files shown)


alexbelgium:
Looks great, thanks very much!

alexbelgium:
The build fails but I don't understand why...

DDanii:
There is something wrong with yamllint in frenck/action-addon-linter, community addons fail too https://github.com/hassio-addons/addon-uptime-kuma/actions/runs/5577908421/jobs/10191346270

alexbelgium:
Nicely seen. Means we'll have to push a new version when this is solved as alas it is a prerequisite step that I prefer to keep on the build workflow

PR #913, opened 2023-07-22
Update gitea

baldarn:
same pr, kind of sad I contribute only in this way :/ but still ;)


alexbelgium:
Contributions are more than welcome! Thanks!

PR #916, opened 2023-07-24
update photoprism

baldarn:
please have a quick check because I'm not 100% sure this will work.


alexbelgium:
As dockerfile is fixed to preview it should work

baldarn:
this edit didn't trigger the update.
do you think that something is missing?

alexbelgium:
Ah indeed, it's when the config.json version is updated that the build is triggered!

PR #947, opened 2023-08-22
Move to CastSponsorSkip

bruvv:
sponsorblockcast has been superseded by CastSponsorSkip, written by gabe565. I recommend moving over to it for improved performance and privacy. sponsorblockcast will no longer be maintained.


bruvv:
Ah good catch on the other stuff thanks @alexbelgium :) (dankjewel)

alexbelgium:
Thanks very much! I've merged it. BTW will it continue working when the same variables and such?

bruvv:
Ah no, sorry working atm will fix that tomorrow!

Issue #734, opened 2023-02-26
✨ [REQUEST] gluetun


kolossboss:
**Which addon?**

[gluetun](https://github.com/qdm12/gluetun)

A docker image to connect to your vpn provider.

**Is your feature request related to a problem? Please describe.**

I would feel better if I can route my addons to a VPN

**Describe the solution you'd like**

A addon with gluetune image. When active and running, you can route other docker images through gluetun, using Portainer to configure.

Thanks a lot

alexbelgium:
Hi, you could use the official addon Tailscale. I'm usually trying to avoid duplicating existing functions, plus this is an official addon which means you'd have a much better support. Also, I really don't have much time to add addons for the moment... Thanks!

kolossboss:
Hi,
Thanks for reply.
Are your sure Tailscale has the same functionality?
With gluetun, you can connect to a paid VPN provider like ExpressVPN or NordVPN.
Tailscale is only to connect your private devices.
Or do I misunderstand something?

I understand if you dont have the time for this. Its just a request. Thanks for all the time you put into this 😁

alexbelgium:
Well I don't use the vpn so I'm not sure if it has the same function... I just remembered having read about tailscale ;-) actually I'm wrong it seems it was wireguard that allowed it (https://www.nstec.com/how-to-set-up-home-assistant-with-a-vpn/)

kolossboss:
Gluetun is more to encrypt internet traffic for "sensitive" data like torrent, jackett or usenet apps (like nzbget).
Hassio Addons like wireguard or tailscale are more to connect to your local network from an extern connection.

alexbelgium:
Ok, thanks for the info. Transmission and qbittorrent have vpn capabilities (especially transmission) but indeed then there is no other existing solution

kolossboss:
Yes, but sadly both are torrent client, I'm more familiar with usenet. 😃 
There is also a usenet newsreader client with openvpn support ([SABnzbd](https://hub.docker.com/r/mumiehub/sabnzbdvpn)).
But the main problem (time) still exists 😁 
(And with gluetun can encrypt every docker container if needed)

lotohov:
let me vote for this addon cos it add vpn to any docker container .. but sadly still this cant be configured from ha.

kolossboss:
You can configure the network this very easy with Portainer. I'm not sure if Home Assistant detect this as "unsupported".
[...]

PR #958, opened 2023-08-25
Initialize Gluetun

BeachyHeadCode:
This is untested but the majority of the environment vars for the docker have been added per the wiki here: https://github.com/qdm12/gluetun-wiki/tree/main#table-of-contents

Unsure if I created all of the dockerfile scripts that are required and removed the ones not needed.

Implementing issue #734

TODO:
- [x] Create files.
- [x] Verify container obtains injected files.
- [ ] Verify injected files function as intended.
- [ ] Release for review


MrPlow254:
@alexbelgium is there a way to test build the existing dockerfile without having it look for a project on this repo? I have a local-addon loaded in HA throwing this error:

`Can't install ghcr.io/alexbelgium/gluetun-armv7:3.35.0: 500 Server Error for http+docker://localhost/v1.42/images/create?tag=3.35.0&fromImage=ghcr.io%2Falexbelgium%2Fgluetun-armv7&platform=linux%2Farm%2Fv7: Internal Server Error ("Head "https://ghcr.io/v2/alexbelgium/gluetun-armv7/manifests/3.35.0": denied")`

Not fully understanding the documentation from HA https://github.com/home-assistant/builder in comparison to how projects are set up on this repo.

alexbelgium:
Hi, thanks! Just delete the image tag in the config.json and it will locally build :) haven't had time to check your work yet thanks!

MrPlow254:
Is it possible to use translations? https://github.com/home-assistant/addons/blob/master/mariadb/translations/en.yaml

I also need to do research on how to override the network mode for the docker services. For example Jackett plugin currently uses port 9117. That needs to be bypassed and a network mode needs to be set to Gluetun to pass all traffic through this addon. Very simple to do with a container file but as you mentioned in #734 HA isn't meant to adjust this way out of the box.

alexbelgium:
Hi, of course, translations work nice. I had removed them as my repo was too big and caused issues, however for a single addon it should be fine

Regarding your point about ports, HA actually allows port mapping in the same way as a composer file : if in the Jackett addon configuration you specify a different port, this will be the one exposed to the outside. The only case when it doesn't work is if the addon is using the host network - the mapping then doesn't work and the original port is forced

alexbelgium:
Would you want me to first the current version in the repo to see how it goes ? We can put a disclaimer in the name (such as DONTUSE) to prevent people from trying it before it is completed and complaining :-)

MrPlow254:
I don't think it is ready yet to complete the pull request.

I will verify the mapping suggestion you made. So essentially what you are saying for example with Jacket:

1. Jacket addon container specifies 9117 internally.
2. HA maps Jacket externally to port 9000 in this example.
3. HA maps Gluetun maps port 9000 externally.
4. Gluten addon container specifies 9117 internally.

I'm unsure if this process then throws the output into a single port to the external internet. I have not tested Gluetun in HA other than getting the installer and config working. I do have Gluetun working in a composer file on the same machine.

MrPlow254:
Well pushed all the config changes and fixed Gluetun not having permissions for networking.

Set up the config file how I want it however i am getting a funky error:
image

The OpenVPN username the environment variable OPENVPN_USER is set yet the container is saying it isn't set. Is there something I am missing? Do we just need to define the environment variable in the config.json file?

MrPlow254:
Well pushed all the config changes and fixed Gluetun not having permissions for networking.

Set up the config file how I want it however i am getting a funky error:
image

The OpenVPN username the environment variable OPENVPN_USER is set yet the container is saying it isn't set. Is there something I am missing? Do we just need to define the environment variable in the config.json file?

alexbelgium:
> I don't think it is ready yet to complete the pull request.
> 
> I will verify the mapping suggestion you made. So essentially what you are saying for example with Jacket:
> 
> 1. Jacket addon container specifies 9117 internally.
> 2. HA maps Jacket externally to port 9000 in this example.
> 3. HA maps Gluetun maps port 9000 externally.
> 4. Gluten addon container specifies 9117 internally.
> 
> I'm unsure if this process then throws the output into a single port to the external internet. I have not tested Gluetun in HA other than getting the installer and config working. I do have Gluetun working in a composer file on the same machine.

Hi, this is exactly how it works, in the end it's just like a docker compose. As long as the host network option is not used in config.json

alexbelgium:
> Well pushed all the config changes and fixed Gluetun not having permissions for networking.
> 
> Set up the config file how I want it however i am getting a funky error: image
> 
> The OpenVPN username the environment variable OPENVPN_USER is set yet the container is saying it isn't set. Is there something I am missing? Do we just need to define the environment variable in the config.json file?
> 
> It is also not creating the /config/addons_config/gluetun folder so i am unable to link that with the containers /gluetun/ directory.

Did you map the config folder in config.json? Here is an example : https://github.com/alexbelgium/hassio-addons/blob/534b65c4635614917530c181bfdc52ddc10cb3f9/baikal/config.json#L58-L62

I've checked and actually the upstream image entrypoint doesn't use /etc/cont-init.d but directly executes a script : https://github.com/qdm12/gluetun/blob/1ac031e78c5cc1732bcab4781ace85939eebdee6/Dockerfile#L202

Either you need to inject your scripts by modifying the file /gluetun-entrypoint (for example you inject your code execution wirh `sed -i "1a /./etc/cont-init.d/20-folders.sh" /gluetun-entrypoint` or more easily in those circumstances I use a custom launch script. If you add this file to your addon : https://github.com/alexbelgium/hassio-addons/blob/master/fireflyiii/rootfs/entrypoint.sh,  modify the dockerfile to launch it : https://github.com/alexbelgium/hassio-addons/blob/534b65c4635614917530c181bfdc52ddc10cb3f9/fireflyiii/Dockerfile#L80-L82, then create a file named 90-run.sh to run the previous script. Your run.sh can be as light as : 
```
#!/bin/bash

/./gluetun-entrypoint
```

Also, variables defined in the addon options are just put in a file named /data/config.json (or something likes that). You need to use to then pass the variables to your script. Either it is done manually, or you can use my global_variables script. To do that, just add `00-global_var.sh` to the modules as here : https://github.com/alexbelgium/hassio-addons/blob/534b65c4635614917530c181bfdc52ddc10cb3f9/fireflyiii/Dockerfile#L54 and it will execute a script that will take all your addon options and pass them as env variables to the app

MrPlow254:
@alexbelgium I am aimlessly attempting to figure this out lol. Adding the CMD /entrypoint.sh prompts in the log it does not exist. So it seems the COPY rootfs/ / does nothing

MrPlow254:
I was able to tinker around with this and verify that the files are getting added into the container. However, The log does not output the correct information and the config file is still not updating the container environment variables.

image

20-folder.sh and 90-run.sh both are functioning correctly themselves.

PR #962, opened 2023-08-26
[Whoogle] Use latest tag

bruvv:
It is now pinned to 0.8.2 and when the bot updates to a newer version it does not update the tag so switching to latest will help since it is now on 0.8.2 instead of the mentioned 0.8.3.


alexbelgium:
Thanks!

PR #964, opened 2023-08-27
[CastSponsorSkip] Small fix for startup and readme

bruvv:
I was getting: `2023/08/27 10:51:25 ERROR Failed to query segments. Retrying... device="Bedroom TV" error="invalid response status: 400 Bad Request"` 
I changed the startup command hopefully that helps.


alexbelgium:
Thanks very much! The redirect of run.sh output was in case the log doesn't show in the add-on log

bruvv:
Ah did not know that, perhaps add that back?

alexbelgium:
Well perhaps it was not needed and it was added by default. let's test it and we'll re-add only if there is no app logs shown in the add-on logs

Issue #967, opened 2023-08-27
🐛 GitHub CI workflow is broken


reedy:
### Description

![Screenshot 2023-08-27 at 15 06 38](https://github.com/alexbelgium/hassio-addons/assets/67615/d6f945f6-7d5c-4234-a442-21afd594d704)

![Screenshot 2023-08-27 at 15 05 43](https://github.com/alexbelgium/hassio-addons/assets/67615/425ac298-e140-4446-9636-c7213d9bffae)

```
Invalid workflow file: .github/workflows/onpr_check-pr.yaml#L97
The workflow is not valid. .github/workflows/onpr_check-pr.yaml (Line: 97, Col: 5): Unexpected value 'uses'
```

### Reproduction steps

```bash
Have a PR merged into this repo...
```

### Addon Logs

```bash
n/a
```

reedy:
![Screenshot 2023-08-27 at 15 08 57](https://github.com/alexbelgium/hassio-addons/assets/67615/70f1da21-d868-4a49-8d51-f6a85a339e28)

reedy:
Not sure if you can just remove the `runs-on` from that section...

reedy:
https://github.com/alexbelgium/hassio-addons/commit/b1be295225cfd6d641958c20d834df276c302104

alexbelgium:
Thanks - I've reverted to the code that I had before wednesday as actually the current logic wouldn't have worked anyway ;-) it would have called the build script in a way that would have committed the changes instead of testing it

PR #968, opened 2023-08-27
onpr_check-pr.yaml: Remove `runs-on` from check-build

reedy:
https://github.com/alexbelgium/hassio-addons/issues/967

### DIFF
--- a/.github/workflows/onpr_check-pr.yaml
+++ b/.github/workflows/onpr_check-pr.yaml
@@ -90,7 +90,7 @@ jobs:
           path: "./${{ matrix.addon }}"
 
   check-build:
-   name: Test addon build
-   if: ${{ needs.check-addon-changes.outputs.changedAddons != '[]' }}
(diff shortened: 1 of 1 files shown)


reedy:
Looks like the job now correctly skips...

Screenshot 2023-08-27 at 15 16 40

reedy:
https://github.com/alexbelgium/hassio-addons/commit/b1be295225cfd6d641958c20d834df276c302104

alexbelgium:
Hi, sorry I hadn't seen your PR

The source of the issue is that a few days ago I wanted to call the build workflow directly but actually it is a bad idea as it would perform the commits instead of just testing the build.

I'll change the logic again then to avoid calling the main build workflow

alexbelgium:
The issue with the previous logic code is that it tried to watch for the config.json/yaml file which is not strictly required as the Dockerfile can build from a BUILD directive directly.

PR #976, opened 2023-09-02
[Castsponserblock] Added CSS_MUTE_ADS config option

bruvv:
New upstream version v0.4.0 of castsponserblock added an option to mute ads. This defaults to true but can be set to false, added that option.


alexbelgium:
Thanks!

PR #977, opened 2023-09-02
[Sponserblock] added only categories as option

bruvv:
Hide the other stuff, is this the way to do it @alexbelgium ? I just wnat the css_categories to be visible by default

### DIFF
--- a/sponsorblockcast/config.json
+++ b/sponsorblockcast/config.json
@@ -14,12 +14,7 @@
   "init": false,
   "name": "Castsponsorskip",
   "options": {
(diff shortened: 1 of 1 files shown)


alexbelgium:
Hi, actually as long as there is a value in `options` it will be visible. And you can't remove it otherwise it will revert to default. So the true solution is to remove all elements from the `options` and only keep them in the `schema` as long as the default is set somewhere

bruvv:
Thanks Alex, removed the extra stuff :)

alexbelgium:
Looks great thanks!

PR #1020, opened 2023-10-06
SSL option for Tandoor Recipes 

blowk:
Testing needed


alexbelgium:
Hi, thanks for your additions.

Here are a few comments : 
## config.json
Regarding config.json : you've added the "schema" elements for ssl. What is missing is that you need to add the ssl folder to the mapping; and https to the url. If you want to access your HA ssl you should add :
```
  "map": [
    "config:rw",
    "share:rw",
    "media:rw",
    "ssl:ro"
  ],
```
and to the webui (actually the whole option seems missing) :
```
  "webui": "[PROTO:ssl]://[HOST]:[PORT:8080]"
```

## mapping ssl to tandoor
According to tandoor documentation, ssl is not supported natively https://docs.tandoor.dev/install/synology/#6-additional-ssl-setup. That means that the way you have started (using nginx) is the right one.
Nginx is already in the dockerfile https://github.com/alexbelgium/hassio-addons/blob/83182e0e5e0a687bdb5230fcea9a1772c76d4bd6/tandoor_recipes/Dockerfile#L45 so that's good.

Now, given that there is a single opened external port (8080), we need to therefore to map tandoor to either use ssl, or not, on this single port, and redirect it to the normal port.

I understand that's the purpose of your nginx code, however I don't recognize the ` {{ if not .ssl }}` command. How will nginx check if ssl is required or not ? In similar occasions I had a script that checked if ssl was required and modified the nginx script as required.

Example for mealie : script to modify is https://github.com/alexbelgium/hassio-addons/blob/83182e0e5e0a687bdb5230fcea9a1772c76d4bd6/mealie/rootfs/etc/cont-init.d/31-nginx.sh ; code for nginx is https://github.com/alexbelgium/hassio-addons/blob/83182e0e5e0a687bdb5230fcea9a1772c76d4bd6/mealie/rootfs/etc/nginx/servers/ssl.conf#L8

Now if 8080 is used as a direct port, the logic would likely be :
- Force tandoor to listen to a different port than 8080, such as 8081
- Have an nginx code that listens on 8080, and routes data to 8081
- Have a script that checks if ssl is enabled in the config.json, and if enabled, adds the ssl elements to the nginx code

alexbelgium:
Additional element : "ssl" : "str?", should be "bool?" to allow true/false statements

Thanks!

alexbelgium:
Let's try that way :-) hopefully nothing will break

alexbelgium:
I've merged it, let's see !

alexbelgium:
Thanks, so should we abandon this? I had reverted the version as it seemed to cause issues and would need additional work

alexbelgium:
What i'll do is push the add-on with the sll code on my test repo so we'll have more flexibility

blowk:
I've put it on your test repo - https://github.com/alexbelgium/hassio-addons-test/pull/4

PR #1042, opened 2023-10-22
update photoprism

baldarn:


baldarn:
I'm not 100% this is going to work, can you check? thanks!

alexbelgium:
Hi, currently it's pulling the "preview" tag so no need to fix a specific tag. The update is done periodically

alexbelgium:
Thanks!

baldarn:
sorry to bother, wouldn't be better to tag to a specific image tag? like this the behavior is unpredictable :/

alexbelgium:
I was thinking that always pulling the "preview" or "latest" tag ensured a consistency in quality ? Rather than individual tags for which it is much more complex to automate

PR #1059, opened 2023-11-02
Upgrade to the latest stable release of NextCloud

ShonP40:
Tested this on a different machine


alexbelgium:
Thanks! Usually automatic updates are done on saturdays otherwise, but yours is the correct way to manually update during the week

ShonP40:
Oh, alright, thanks!

PR #1094, opened 2023-11-24
docs(readarr): update README.md with clearer config instructions

### LINKED ISSUE #1088: 🐛 [readarr] Update info page with better information
### Description

> Configurations can be done through the app webUI, except for the following options

image

so how do we (as homeassistant users) do that configuration? 

perhaps this info page for the addon should then go on to say something like: 

- Using an editor on the homeassisntant server (like Studio Code Server), edit the `/config/addons_config/readarr/config.xml` ? 

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Addon Logs

```bash
not relevant
```

### Architecture

amd64

### OS

HAos

airtonix:
fixes #1088

This changes makes it clearer as to where and what can be done with configuration for readarr


alexbelgium:
Thanks very much! I don't have much time to support the add-ons; so documentation is the point where I also appreciate the most contributions ;)

alexbelgium:
Actually there was link to the wiki both in the readme and the log describing the usage of the config.yaml file but it was incomplete. I've updated the instructions https://github.com/alexbelgium/hassio-addons/wiki/Add%E2%80%90ons-feature-:-add-env-variables

airtonix:
Yep i made sure that the link was still included, but it's important from a time poor users perspective that you give them the salient details asap and as plainly as possible (ie don't leave them guessing).

PR #1133, opened 2023-12-14
update gitea to 1.21.2

baldarn:
what do you think to create an automatic checker for this kind of updates?


alexbelgium:
Hi, OK ! I can implement automatic updates based on upstream changes. Only constraints is that it would run once per week on Saturday to avoid having too many chances for add-ons with daily updates

alexbelgium:
Thanks!

Issue #1172, opened 2024-01-12
🐛 [Qbittorrent] Web UI artifacts


akrigator:
### Description

<img width="1540" alt="image" src="https://github.com/alexbelgium/hassio-addons/assets/695335/cbe8e3b3-7217-4a1d-95c9-df00f9b82b4b">

### Reproduction steps

qbittorrent/rootfs/etc/cont-init.d/91-qbittorrent_configuration.sh line 193:

inner request `curl -f -s https://api.github.com/repos/WDaan/VueTorrent/releases` get response  

```
{
  "message": "Moved Permanently",
  "url": "https://api.github.com/repositories/181508472/releases",
  "documentation_url": "https://docs.github.com/rest/guides/best-practices-for-using-the-rest-api#follow-redirects"
}
```

### Addon Logs

```bash
[08:11:58] WARNING: Default Webui selected ! It will not work for ingress, which will stay with vuetorrent

curl: (3) URL using bad/illegal format or missing URL
Error : /etc/cont-init.d/91-qbittorrent_configuration.sh exiting 3
/etc/cont-init.d/92-qbit_manage.sh: executing
/etc/cont-init.d/93-openvpn.sh: executing
[08:11:59] INFO: Direct connection without VPN enabled
```

### Architecture

armv7

### OS

HAos

akrigator:
I guess that that fix is easy, just to add `-L` option to follow redirection:
```
curl -f -s -L https://api.github.com/repos/WDaan/VueTorrent/releases
```

alexbelgium:
Thanks very much for the PR!

alexbelgium:
Btw, watch-out that linuxserver does not anymore updates their armv7 images ; if you have the opportunity it would highly be recommended that you switch to aarch64 as we would have more up-to-date addons

akrigator:
Actually it was aarm64, my bad

akrigator:
Fixed

PR #1173, opened 2024-01-12
Update 91-qbittorrent_configuration.sh

akrigator:
Fix qBittorrent WebUI #1172


alexbelgium:
Thanks very much! I'll add an information message in case of failure for future reference in the code

PR #1185, opened 2024-01-15
Fix: Flaresolverr healthcheck path

override80:
Since flaresolverr has a specific endpoint for healthchecks (/health) it could make sense to use it insted of the root. Also, this will not flood the log when using INFO LOG_LEVEL (default) with entries like:

`2024-01-15T17:49:56.962520+01:00 hassio addon_db21ed7f_flaresolverr[816]: 2024-01-15 17:49:56 INFO     127.0.0.1 GET http://127.0.0.1:8191/ 200 OK`

### DIFF
--- a/flaresolverr/CHANGELOG.md
+++ b/flaresolverr/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Correct healthcheck endpoint (thanks @override80)
 
 ## 3.3.13 (13-01-2024)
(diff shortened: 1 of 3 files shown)


alexbelgium:
Thanks very much! The only remaining element is to increment the version in config.json either your add it to the PR or I'll do it a bit later

PR #1191, opened 2024-01-21
Update docker hub from omni-nightly to latest nightly

JohnnyPicnic:
Mealie is now stable. Latest docker hub is just nightly

### DIFF
--- a/mealie/build.json
+++ b/mealie/build.json
@@ -1,7 +1,7 @@
 {
   "build_from": {
-    "aarch64": "hkotel/mealie:omni-nightly",
-    "amd64": "hkotel/mealie:omni-nightly"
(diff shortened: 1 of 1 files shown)


alexbelgium:
Thanks very much! The switch from images is seamless, the underlying systems are exactly the same?

JohnnyPicnic:
Previous version was v1.0.0-RC1.1. New version is v1.0.0. I was able to migrate seamlessly using amd64 images. 
Does your updater create a new image automatically from docker hub?

alexbelgium:
Well I'll push it now! Automatic updates ara every Saturdays. What actually triggers the build is the change of version in config.yaml

PR #1195, opened 2024-01-22
feat: add media mount for portainer

nkm8:
I use portainer to run Channels DVR, which requires access to my mounted media folders.

### DIFF
--- a/portainer/config.json
+++ b/portainer/config.json
@@ -19,6 +19,7 @@
   "init": false,
   "map": [
     "share:rw",
+    "media:rw",
     "ssl"
   ],
   "name": "Portainer",


alexbelgium:
Hi, are you sure it works that way? Shouldn't you map the media mount in the container that you are starting? By mapping /media here, it will be accessible from within the portainer container but not from within the others containers that you could start with it. Thanks for the confirmation!

nkm8:
@alexbelgium Thank you for the explanation! I hadn't tested because I assumed `/media` wouldn't be available to mount in portainer containers without this change. I confirmed that you are correct and this change isn't necessary - please disregard.

PR #1280, opened 2024-03-02
Fix cron problem

MiguelAngelLV:
The cron scheduler maybe don't implement all cron options and does not work fine with 0/6.

With 0/6 only execute at 0 and at 6, so, you need relogin every day (the token need refresh each 8 hours).

Changing 0/6 to */6 it works fine.


alexbelgium:
Thanks!

PR #1288, opened 2024-03-05
update: add ports to omada config

ttocsr:
There are additional ports needed in the latest controller version.
This is needed for things like DPI to work.


alexbelgium:
Thanks!

PR #1364, opened 2024-04-29
WHATSAPPER | Fix error in docker build

baldarn:
fix docker build


baldarn:
@alexbelgium I removed the line that made fail the build. this seems to work. we can try to merge in order to see if everthing is ok now :)

PR #1387, opened 2024-05-09
OPENPROJECT | Add openproject

baldarn:
I have the project of add openproject, so I added openproject to the project

XD


baldarn:
still WIP.

will have to test this and volumes configurations ;)

baldarn:
@alexbelgium I can't understand why `Codacy Static Code Analysis` is failing. why is that?

alexbelgium:
Really nice thanks! I'm off for a couple days and will gladly help when I'll be back if needed!

alexbelgium:
> @alexbelgium I can't understand why `Codacy Static Code Analysis` is failing. why is that?

I think codacy is 

> @alexbelgium I can't understand why `Codacy Static Code Analysis` is failing. why is that?

Actually it is useless things that I have white listed on my own codacy but it seems it doesn't transfer to your checks... So you can just ignore it :)

baldarn:
the sad thing is that I can't try that. could we try and merge this to see if this is working?

alexbelgium:
Oh sorry I missed your message i'll merge right now!!

And thanks very much for the coffees on PayPal!

alexbelgium:
Hi! I've tested it but you miss an instruction to start the actual app : you should have a file in rootfs/etc/cont-init.d/99-run.dh that would contain the startup code 

For the moment it just builds and executes the basic scripts but the app doesn't start. The best way to see what to put in the run.sh file is to check the entrypoint or cmd of the initial dockerfile

alexbelgium:
Here is the command "./docker/prod/entrypoint.sh" taken from https://hub.docker.com/layers/openproject/openproject/dev/images/sha256-3eeb74aa73c6191c397c9c31c77176f350dc6d0722551dbd8fa3ad964ebbb8a2?context=explore

alexbelgium:
It actually is likely "exec /app/docker/prod/supervisord /app/docker/prod/entrypoint.sh" but then it says an error about file but found while starting

alexbelgium:
You can troubleshoot and test by adding in the sh file of your /config either sleep infinity and then test commands from portainer console, or

```
#!/usr/bin/env bashio
# shellcheck shell=bash

echo "Starting"
#find / -name entrypoint.sh 2>/dev/null
#./docker/prod/entrypoint.sh

cp /app/docker/prod/supervisord /config/a
cp /app/docker/prod/entrypoint.sh /config/b

/./app/docker/prod/supervisord /app/docker/prod/entrypoint.sh
```

baldarn:
what if I remove the `ENTRYPOINT [ "/ha_entrypoint.sh" ]`?

what are the pro and cons?

alexbelgium:
it all depends how is built the upstream container : 
- if it uses s6 supervision, then the line "ENV S6_STAGE2_HOOK=/ha_entrypoint.sh" will still make it execute
- if it uses a simple system, then you can indeed use my custom entrypoint, and reference the previous entrypoing in a /etc/cont-init.d/run.sh file as I described
- if the addon really needs the initial entrypoint to boot, then you can remove the line (as you wrote above) but execute the ha_entrypoint.sh as part of the initial entrypoint by using sed. Here is an example : 
https://github.com/alexbelgium/hassio-addons/blob/4759ce5431f92024eeb8edcef8ff909271d13366/wger/Dockerfile#L39-L40

alexbelgium:
btw if you want to launch yourself the building of a new addon version you can just open a new pull request with the changes and write `/automerge` in the comments

baldarn:
I will try with the third option and use the magic `/automerge` to see if something starts ;)

PR #1408, opened 2024-05-22
OPENPROJECT | Update to 14.1.0

baldarn:
Update to 14.1.0


baldarn:
@alexbelgium I think is me, but the docker image is not build
Screenshot_20240522_111944

maybe with the automerge this is not triggered?

alexbelgium:
Build is not completed yet? You can check in the Actions if it was completef

baldarn:
mmmm I think that wasn't triggered

when should that be triggered? on merge

baldarn:
@alexbelgium how did you fix this? so next time I'll try to do without messing with you ;)

alexbelgium:
hehe I forced the image update by changing the addon version number :-) I didn't check but we can see at your next automerge if it starts the script correctly :-)

PR #1411, opened 2024-05-22
NAVIDROME | Fix navidrome image

baldarn:
my bad... to fast ctrl c, ctrl v


alexbelgium:
if you want a nice icon you need to use a png with transparent background ; it will allow it to be colored when the addon is active ; grayed when the addon is inactive

baldarn:
uuuu nice, didn't know

i see that home assistant apply a shade. but good to know ;)

PR #1455, opened 2024-06-27
changes to vlc instructions

thor0215:
I forgot to mention that birdnet-go will need to connect to the vlc rtsp stream using udp. I updated the readme to reflect that.  I also simplified the vlc --sout command line parameter.


alexbelgium:
Thanks very much!

PR #1484, opened 2024-07-21
Add Authentik SSO

ddcash:
Added the option to support using Authentik SSO.  Confirmed working following instruction from https://docs.linkwarden.app/self-hosting/sso-oauth


alexbelgium:
Thanks!

PR #1486, opened 2024-07-21
Add option to disable registration

ddcash:
Add the options to disable registration and if using SSO the option to disable the username and password fields

using the variables from https://docs.linkwarden.app/self-hosting/environment-variables


alexbelgium:
I've incremented the version

Thanks!

PR #1488, opened 2024-07-21
Update config.json

ddcash:
I missed modifying the schema and in addition changed the enable/disable options to use "bool" instead of "str" to align better with the options


ddcash:
Sorry for the slight oversights and multiple Pulls.  I'm fairly new to all this and appreciate all you do.

alexbelgium:
Thanks! No worry I'm on holiday so don't review much also ;) but I learned a lot through trial and errors so fully understand that you follow the same path

PR #1493, opened 2024-07-24
[ChangeDetect.io] Add instructions on how to create sidebar shortcut

### LINKED ISSUE #1492: ✨ [ChangeDetection.io] Any chance it's possible to automatically create a sidebar entry?
### Checked

- [X] If a new addon, I have checked on  Google that such as addon doesn't already exists
- [X] If a new feature, I have searched in the discussions & issues that it was not already covered

### Which addon?

ChangeDetection.io

This add-on (from the very few I have) is a bit odd to access - when you press "Open Web UI", it opens a new tab at the server URL, instead of opening it as an iframe of sorts inside the HA interface.

I understand it's because the add-on doesn't use Ingress to embed itself into HA - as, for instance, the official Grocy add-on does.

However, from _my_ point of view, Ingress would not be needed if it's complex to implement (although it would probably increase the add-on rating).

I have a  simpler suggestion, as I don't think a sidebar entry could be automatically created without Ingress: I can submit a small PR to its README explaining how to create a custom dashboard that embed the resulting interface onto HA via an iframe - Settings > Dashboards > Add Dashboard > Webpage.

image

Not sure if you think this could be useful on other add-ons.

igorsantos07:
close #1492

Feel free to copy over / adapt those instructions to whatever else add-ons this could be useful :)

I don't use any other add-on, so no clue where it could be handy or not.

> Also tried to fix the stargazers image, and did some tiny fixes to the text that I stumbled upon :sweat_smile:

### DIFF
--- a/changedetection.io/README.md
+++ b/changedetection.io/README.md
@@ -17,7 +17,7 @@
 
 _Thanks to everyone having starred my repo! To star it click on the image below, then it will be on top right. Thanks!_
 
-[![Stargazers repo roster for @alexbelgium/hassio-addons](https://raw.githubusercontent.com/alexbelgium/hassio-addons/master/.github/stars2.svg)](https://github.com/alexbelgium/hassio-addons/stargazers)
+[![Stargazers repo roster for @alexbelgium/hassio-addons](https://reporoster.com/stars/alexbelgium/hassio-addons)](https://github.com/alexbelgium/hassio-addons/stargazers)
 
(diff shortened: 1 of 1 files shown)


alexbelgium:
thanks, I've also made a wiki out of it : https://github.com/alexbelgium/hassio-addons/wiki/Add-to-sidebar-with-iframe-@igorsantos07

PR #1538, opened 2024-08-19
Update Config Location

kool2zero:
Fix misspelling in config location.

image

### DIFF
--- a/.github/workflows/onpush_builder.yaml
+++ b/.github/workflows/onpush_builder.yaml
@@ -16,7 +16,7 @@ on:
 
 jobs:
   correct_path_filters:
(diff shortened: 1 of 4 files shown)


alexbelgium:
Hi, thanks we'll also need to transfer files for users : check if the previous location had files in it and transfer it to the corrected folder name then remove the previous folder name 

Also you'll need to Increment the version in config.yaml ( I use -2 usually) for a new build to launch. Thanks!

kool2zero:
I pushed the updated code based on your feedback,

alexbelgium:
Thanks!

alexbelgium:
Really nice

PR #1543, opened 2024-08-22
Added missing USB device types

eroldan:
There were some smartctl's device types missing. Added the "snt*" types, which I use. Information taken from https://www.smartmontools.org/wiki/USB


alexbelgium:
Thanks!

PR #1593, opened 2024-10-03
Update HAINTEGRATION.md

thor0215:
Made a change to BirdNET-Go Events Sensor. The sensor will now work if you have other triggers defined for other sensors not related to Birdnet


alexbelgium:
Thanks very much!

PR #1662, opened 2024-12-10
Add discovery to Mealie config

andrew-codechimp:
Integrations can support discovery of add-ons if the add-on supports it.

Adding discovery for Mealie so that the integration can detect the add-on and suggest a setup.  Note this is not in place within the integration at the moment as needs the discovery tag into the add-on to detect this, I am not an add-on dev so have been unable to work out where to add this into my own config to test.

See adguard example of where this has been added.
https://github.com/hassio-addons/addon-adguard-home/blob/f7d3615c616d942ef150e7bc0afc5a1c9e4060bc/adguard/config.yaml#L25

### DIFF
--- a/mealie/config.json
+++ b/mealie/config.json
@@ -113,5 +113,5 @@
   "slug": "mealie",
   "udev": true,
   "url": "https://github.com/alexbelgium/hassio-addons",
-  "version": "v2.3.0"
+  "version": "v2.3.0-test"
 }


alexbelgium:
Hi, thanks actually this is an undocumented feature :
image https://developers.home-assistant.io/docs/add-ons/configuration/

I will still push it in a test version and we'll see!

andrew-codechimp:
Yes, it seems the docs need updating as I've come across a few that do this.
Once the test version is available I'll work on the integration side to see if I can get it working.

alexbelgium:
Thanks for the pointer!

joostlek:
I checked with Stefan from the supervisor, and apparently we also need a script that publishes the data to the integration. Adguard has that located here https://github.com/hassio-addons/addon-adguard-home/tree/main/adguard/rootfs/etc/s6-overlay/s6-rc.d/discovery

andrew-codechimp:
Looking at the history of adguard there is an older discovery implementation that follows the cont-init.d pattern used here, not sure if this could be dropped in and just the `service adguard` reference changed.

https://github.com/hassio-addons/addon-adguard-home/blob/f6dbb7dbdece33d58ea298dac2af06853505e1c8/adguard/rootfs/etc/cont-init.d/discovery.sh

Issue #1655, opened 2024-12-05
🐛FREE GAMES CLAIMER:  Fails to load credentials and run. 


codeclinic:
### Description

Even though I have added my credentials in the yaml file, I get this log output everytime, just looping.....
Starting...
/etc/cont-init.d/00-banner.sh: executing
Add-on: Free Games Claimer (NoVNC not working)
automatically claims free games on the Epic Games Store, Amazon Prime Gaming and GOG
Add-on version: 1.6
You are running the latest version of this add-on.
System: Home Assistant OS 14.0 (aarch64 / raspberrypi5-64)
Home Assistant Core: 2024.12.0
Home Assistant Supervisor: 2024.11.4
Please, share the above information when looking for help
or support in, e.g., GitHub, forums
Provided by: https://github.com/alexbelgium/hassio-addons
/etc/cont-init.d/00-global_var.sh: executing
CMD_ARGUMENTS='node epic-games ; node prime-gaming ; node gog'
CONFIG_LOCATION='/config/addons_config/free_games_claimer/config.env'
Timezone set from Etc/UTC to Europe/Berlin
/etc/cont-init.d/01-config_yaml.sh: executing
Setting permissions for the config.yaml directory
Load environment variables from /config/addons_config/free_games_claimer/config.env/config.yaml if existing
If accessing the file with filebrowser it should be mapped to /homeassistant/addons_config/free-games-claimer/config.yaml
Wiki here on how to use : github.com/alexbelgium/hassio-addons/wiki/Add‐ons-feature-:-add-env-variables
EG_EMAIL='xxxxxxxxxxxxxxxxxx'
EG_PASSWORD='xxxxxxxxxxxxxxxxxxxxxxxxxxxx'
/etc/cont-init.d/01-custom_script.sh: executing
Execute /config/addons_autoscripts/free-games-claimer.sh if existing

If accessing the file with filebrowser it should be mapped to /config/addons_autoscripts/free-games-claimer.sh
Wiki here : github.com/alexbelgium/hassio-addons/wiki/Add-ons-feature-:-customisation
... no script found, exiting
/etc/cont-init.d/20-folders.sh: executing
Creating config location ...
Copying files if needed...
/etc/cont-init.d/99-run.sh: executing
[15:04:39] WARNING: A default config.env file was copied in /config/addons_config/free_games_claimer. Please customize according to https://github.com/vogler/free-games-claimer/tree/main#configuration--options and restart the add-on
[...]

PR #1689, opened 2024-12-29
fix: resolve config.env directory issue in free_games_claimer add-on (#1655)

ayushtiwari134:
Steps taken:
1. Added a check to detect if config.env is a directory instead of a file.  
2. Removed the invalid directory if found and replaced it with a valid config.env file.  
3. Set proper permissions for the config.env file.  
4. Improved logging for debugging and status updates.


alexbelgium:
Thanks very much! The version is being build

PR #1695, opened 2025-01-05
[New add on] NetAlertX

jokob-sk:
Hi @alexbelgium 👋,

I've tried to create a NetAlertX addon, a project I maintain and develop. Not sure if the submission is correct, I probably missed a few things. I based it off the piwigo install as per https://github.com/alexbelgium/hassio-addons/wiki/Building-an-addon

The docker image I maintain can be found here: https://hub.docker.com/r/jokobsk/netalertx

Ideally the container runs on the host network - not sure if that is possible. If not, that's fine too, just a different set of plugins needs to be enabled by the user to get their network scanned. 

Let me know if I can help further.

Thanks for considering my contribution,
j


alexbelgium:
Hi, very interesting addon! if you want to use host network access, you need to add `host_network: true` in your addon config

alexbelgium:
I've already merged the addon if you want to test, we can increment the versions after to test the different builds. That's what I usually do

jokob-sk:
Thanks a lot @alexbelgium 🙏 I will have a look and report back - also I will check how to add host_network: true iin teh addon config

jokob-sk:
Hummm, can't find it- I probably miss-configured something: 
image

DO you have any tips on how to troubleshoot?

alexbelgium:
Hi, potential issues are usually seen in the PR check prior to merging. I guess I was a bit too fast :-)
You should also see it from the "Checks" tab at the top of the screen, where is shows a couple Mardown issues (https://github.com/alexbelgium/hassio-addons/actions/runs/12615666281/job/35161437159?pr=1695) but no big issues.

The best possible source of info is the build workflow : https://github.com/alexbelgium/hassio-addons/actions/runs/12628632476

It will show if there is any issues with the config file itself, and in the build

alexbelgium:
So, the core issue is that `armv6` is not valid according to https://developers.home-assistant.io/docs/add-ons/configuration/#required-configuration-options

To add `host_network: true` you just need to add it in the config file as for : https://github.com/alexbelgium/hassio-addons/blob/99c2400f880aba213d7249bd295678aec021d2c5/plex/config.json#L117

jokob-sk:
thanks @alexbelgium , new PR fixing these things: https://github.com/alexbelgium/hassio-addons/pull/1696

PR #1705, opened 2025-01-08
[netalertx] another try

jokob-sk:
Hi @alexbelgium ,

Another try - can you have a look at the rootfs folder if I'm doing things correctly before merging?

https://github.com/jokob-sk/hassio-addons/blob/master/netalertx/rootfs/etc/s6-overlay/s6-rc.d/init-alexbelgium/99-run.sh


alexbelgium:
Btw I didn't take the opportunity yet but congratulation for your superb netalertx app! both in function and in visuals it is amazing!

alexbelgium:
In HA there is 2 main persistent directories : /config (accessible) and /data (but accessible). The best could be therefore to replace with grep/sed /app/db with /config/db and /app/config why /config/config? Sometimes I use symlinks instead of grep/sed but it can cause issues for example if the db is sqlite

jokob-sk:
Thanks for the info!

Hummm, replacing the location of the db and config in all the code would require a lot of work. The app in the UI allows you to export some of the configuration, so that could be a workaround which I could mention in the README. 

I'm a symlink noob so is there a way how we could solve this with symlinks? 

Unfortunately the DB is SQLITE but for an almost full restore (although manual) only the content of the `/app/config` folder is required. 

I think for v0.1 if we have a persistent /config folder then that should be good enough + instructions on how to restore their devices: https://github.com/jokob-sk/NetAlertX/blob/main/docs/BACKUPS.md

What do you think?

image
image

alexbelgium:
You could perhaps try something like that : so any hard links are changed 
```
# In the dockerfile Changing hard links if possible
RUN for folder in config db; do \
    echo "Adapting hard links" && \
    grep -rl "/app/$folder" / 2>/dev/null | xargs sed -i "s|/app/$folder|/config/$folder|g"; \
done

# In the addon script, make symlinks on the fly
for folder in config db; do
    echo "Creating for $folder"
    # Create symlinks
    mkdir -p /config/"$folder"
    if [ -d /app/"$folder" ] && [ "$(ls -A /app/"$folder")" ]; then
        cp -rn /app/"$folder"/* /config/"$folder"/
    fi
    rm -r /app/"$folder"
    ln -sf /config/"$folder" /app/"$folder"
done
```

jokob-sk:
Hi @alexbelgium ,

Thanks a lot!

I don't think this here will work due to instances such as:

image
```bash
# In the dockerfile Changing hard links if possible
RUN for folder in config db; do \
    echo "Adapting hard links" && \
    grep -rl "/app/$folder" / 2>/dev/null | xargs sed -i "s|/app/$folder|/config/$folder|g"; \
done
```

I pushed the approach with symlinks to test if possible - if not I will have to refactor the app to make the paths easier to replace. Can you check if the latest commit is correct? Can we try this first before I spend time refactoring?

Thanks in advance!

alexbelgium:
Thanks let's try! Once it is built we can play with  it by force sleeping the script by adding "sleep infinity" in /config/xxx-netalertx and playing with options manually from the command line within the container (using portainer) to see what works and if everything is correctly configured :) I've usually tried this empirical approach with some success as all add-ons can have specific behavioral differences!

alexbelgium:
In you example we could then simply using `sed -i "/config_file/c\\$config_file = /config/config/app.conf`. The c command in sed replaces the whole line. Sorry my help is kind of patchy, I can look a bit deeper in a couple days probably

jokob-sk:
Thanks a lot for the ongoing help - I tried reading the HA docs on plugin creation, but my brain just wouldn't understand it, so you are a massive help 

> In you example we could then simply using sed -i "/config_file/c\\$config_file = /config/config/app.conf. The c command in sed replaces the whole line. Sorry my help is kind of patchy, I can look a bit deeper in a couple days probably

Nah, I understand, everyone has lives outside of FOSS lol. There is also the issue of reverse proxies - not sure if that influences it - although probably we could say reverse proxies not supported - deploy directly if needed. 

I'd rather then spend proper time refactoring this - I don't like how there are so many places where the location is specified.

alexbelgium:
Well easiest is to test if symlinks work (easiest) or else we could also change the WORKDIR so that the relative paths work! Or sed as the last option ;)

alexbelgium:
I've tried it, looks like you are missing the type file in the s6 structure `s6-rc-compile: fatal: unable to read /etc/s6-overlay/s6-rc.d/init-alexbelgium/type: No such file or directory` like https://github.com/alexbelgium/hassio-addons/blob/master/piwigo/rootfs/etc/s6-overlay/s6-rc.d/init-alexbelgium/type

jokob-sk:
re added it in the latest PR, can you have a quick browse - I also removed the `up` file previously - is it needed? https://github.com/alexbelgium/hassio-addons/pull/1713

PR #1713, opened 2025-01-10
[netalertx] re-adding missing files

jokob-sk:
- re-added run - oneshot


jokob-sk:
Do I also need the `up` file?

alexbelgium:
Yes up is required too :)

alexbelgium:
I've pushed a v8 with it and the log seems to boot quite nice!

jokob-sk:
Oh cool, will check when I get back to the pc. On the phone. Should I re add the up file or did you do it?

jokob-sk:
Saw you added the `up` yourself - thank you.

Yes, seems to start up fine 🥳

I quickly added a couple of Export/Import backup options to the app UI in case we have issues with the symlink approach. It's manual, but better than nothing. https://github.com/jokob-sk/NetAlertX/blob/main/docs/BACKUPS.md#-backup-strategies

I will now play around with the symlink setup since the app is running

PR #1765, opened 2025-02-13
[TOR] http tunnel capability

akrigator:
I like to use tor as a  local proxy for application to bypass some restrictions, but not all of them support SOCKS protocol via 9050 port. Let's open 9080 port to use the HTTP tunnel capability, which is build in to the tor to achieve that.

### DIFF
--- a/tor/CHANGELOG.md
+++ b/tor/CHANGELOG.md
@@ -13,3 +13,7 @@
 ## 5.0.1-4 (14-08-2024)
 
 - Add execution permission for permission for /etc/s6-overlay/s6-overlay/s6-rc.d/init-tor/run file
+
+## 5.0.2-1 (13-02-2025)
+
+- HTTP tunneling
--- a/tor/Dockerfile
+++ b/tor/Dockerfile
@@ -1,4 +1,4 @@
-ARG BUILD_FROM=ghcr.io/hassio-addons/base:16.2.0
+ARG BUILD_FROM=ghcr.io/hassio-addons/base:17.1.3
 # hadolint ignore=DL3006
 FROM ${BUILD_FROM}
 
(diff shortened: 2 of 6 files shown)


alexbelgium:
Thanks for the PR ! Looks like armv7 doesn't build, but we could probably just remove it

For my info, why did you dowgrade from hassio-addons/base:16.2.0 to hassio-addons/base:15.0.8 ? It might be good to add a note to avoid automatic updates in the future, thanks!

akrigator:
i'm a bit confused with dependencies )
the last release of tor 5.0.1 from comunity repo has been based on hassio-addons/base:15.0.8 https://github.com/hassio-addons/addon-tor/blob/v5.0.1/tor/Dockerfile
there is also open mr https://github.com/hassio-addons/addon-tor/blob/renovate/alpine_3_21-openssl-3.x/tor/Dockerfile
but i'm failing to build any with packages from https://pkgs.alpinelinux.org/packages?name=libcap&branch=v3.21&repo=&arch=armv7&origin=&flagged=&maintainer=

will try later again
i've thought that armv7 for raspberry p4 which i'm currently using, but it's aarch64 actually. So yeah may be you are right

akrigator:
The build failures root cause in `build_from` - which is defined in 2 places `Dockerfile` and `build.json`

alexbelgium:
Looks like it builds fine though - even if lint fails

akrigator:
Not this time)

akrigator:
@alexbelgium  what does CACHED  mean?
```
#14 [ 9/14] RUN go get     && go build -o /usr/local/bin/snowflake
#14 CACHED
```
This step has been passed on push https://github.com/alexbelgium/hassio-addons/actions/runs/13309403504/job/37167925446 but failed on merge [build](https://github.com/alexbelgium/hassio-addons/actions/runs/13309515916/job/37168356406) 
Does it mean that it was mocked at the run, and really started only at second run? so and problem is really in `go get` or `go build -o /usr/local/bin/snowflake`

alexbelgium:
Copilot says: 

The job failed due to a segmentation fault during the Go build process. This can be caused by various issues, including hardware limitations or bugs in the Go compiler.

Here are a few steps to troubleshoot and potentially resolve the issue:

1. **Update Go Version**: Ensure you are using the latest stable version of Go in your Dockerfile.
2. **Check for Memory Limits**: If you're running on limited hardware, it might help to increase the available memory or use a larger runner.

Here's how to update the Go version in your Dockerfile:

```dockerfile
# Install Go
RUN wget https://dl.google.com/go/go1.18.3.linux-arm64.tar.gz
RUN tar -C /usr/local -xzf go1.18.3.linux-arm64.tar.gz
ENV PATH=$PATH:/usr/local/go/bin

# Add the rest of your build steps here
RUN go get && go build -o /usr/local/bin/snowflake
```

Make sure to adjust the Go version to the latest stable release. Rebuild your Docker image and rerun the workflow.

PR #1769, opened 2025-02-13
[TOR] Init build

akrigator:
one more time

### DIFF
--- a/tor/CHANGELOG.md
+++ b/tor/CHANGELOG.md
@@ -26,3 +26,7 @@
 ## 5.0.2-3 (13-02-2025)
 
 - Fix network for http tunel
+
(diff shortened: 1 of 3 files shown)


akrigator:
ok, i've figured out what a hell is going on) the addon is based on https://github.com/hassio-addons/addon-base/blob/main/base/Dockerfile container which has preinstalled lower version of `musl-utils=1.2.5-r8` and `musl=1.2.5-r8` for 17.1.4 release whenever alpina 3.21 provides `1.2.5-r9`

alexbelgium:
Ahah docker containers are always so complex with their many layers. Hopefully when it will be set it will be ok for a long time

Issue #1755, opened 2025-02-05
🐛 [Transmission] If defined user and pass, the container start longer and report unhealthy


myhomeiot:
### Description

Config example:
```yaml
user: admin
pass: 123456
```

### Reproduction steps

```bash
1. Go to configuration
2. Stop add-on
3. Add to configuration user: admin, pass: 123456
4. Start add-on
5. Check unhealthy of container
```

### Addon Logs

```bash
No log
```

### Architecture

amd64

myhomeiot:
Fixed in `4.0.6-r0-ls272-4` by #1780

PR #1780, opened 2025-02-21
Fix user/pass (if defined) to healthcheck

myhomeiot:
Healthcheck curl doesn't have access to user and pass env variables without source ~/.bashrc.
I also think that more reasonable to change healthcheck interval from 5s to 1m.

Thanks for your great job!

PS: I test all cases locally and can confirm that this time everything works as expected.

## Summary by CodeRabbit

- **Chores**
	- Refined system health monitoring by extending the check interval and updating the initialization process to better account for environment settings, enhancing overall stability and reliability.


myhomeiot:
Related to #1755

PR #1871, opened 2025-05-19
Update config.json

ash286:
Calibre-web can not access `/addon_configs` folder. Without that `metadata.db` is not accessible from Calibre config folder.

I belive that the correct location is `/addon_configs`, not `/addon_config`. Please see below output from the current Calibre-web docker

```
root@a0d7ed7f-calibre-web:/homeassistant/addons_config# ls -la
total 32
drwxrwxrwx  8 root     root     4096 May 13 23:12 .
drwxrwxrwx 22 root     root     4096 May 19 02:23 ..
lrwxrwxrwx  1 root     root       35 May 13 23:12 a0d7ed7f_calibre-web -> /addon_configs/a0d7ed7f_calibre-web
lrwxrwxrwx  1 root     root       31 May 13 23:12 a0d7ed7f_calibre -> /addon_configs/a0d7ed7f_calibre
...
...
...
root@a0d7ed7f-calibre-web:/homeassistant/addons_config# cd a0d7ed7f_calibre
bash: cd: a0d7ed7f_calibre: No such file or directory
root@a0d7ed7f-calibre-web:/homeassistant/addons_config#

```

### DIFF
--- a/calibre_web/config.json
+++ b/calibre_web/config.json
@@ -84,7 +84,7 @@
   "map": [
     "media:rw",
     "share:rw",
-    "addon_config:rw",
+    "addon_configs:rw",
     "homeassistant_config:rw",
     "ssl"
   ],
--- a/calibre_web/config.json
+++ b/calibre_web/config.json
@@ -84,7 +84,7 @@
   "map": [
     "media:rw",
     "share:rw",
-    "addon_configs:rw",
+    "all_addon_configs:rw",
     "homeassistant_config:rw",
     "ssl"
   ],


alexbelgium:
Hi, why should it access /addon_configs ? Normally it should access /config only according to the migration script

Your proposal will not work, it maps the homeassistant folder and not the previous one that I was using. If the path is blocked in your database I can add a symlink to solve that

alexbelgium:
What is your startup log ?

ash286:
> Hi, why should it access /addon_configs ? Normally it should access /config only according to the migration script
> 
> Your proposal will not work, it maps the homeassistant folder and not the previous one that I was using. If the path is blocked in your database I can add a symlink to solve that

The reason it needs access to /addon_configs so that it can get access to metadata.db of Calibre addon. Previously it was working fine with current config.json settings: "addon_config:rw". Now Home Assistant has got update to this [New addon_configs folder](https://developers.home-assistant.io/blog/2023/11/06/public-addon-config/#new-addon_configs-folder), this needs to update with `all_addon_configs:rw`. I'm not requesting any new permission but it's update of settings to reinstate /addon_configs access.

I think this is only way to get access to metadata.db and all books access from calibre addon to calibre-web addon.

ash286:
Agree, transferring metadata to /share would be more complex and it'll need to update calibre addon as well. Thanks for merging my pull request :)

alexbelgium:
Thanks to you. If you have some time would you be able to adapt the readme too ?

alexbelgium:
Hi ok new version available

ash286:
Brilliant, calibre-web now works as expected :)

In a day or two I'll adapt the readme.

PR #1883, opened 2025-06-02
Update 99-run.sh

ash286:
If moving pgdata is not important in order enable the config.yaml env exports, we'd let users to decide attachments location by specifying OPENPROJECT_ATTACHMENTS__STORAGE__PATH.


alexbelgium:
Hi, what is the purpose and benefit for the user ? I usually avoid complexifying addons except when there is a clear need. Also, code must take into account previous storage locations to allow migration and avoid data loss for other users

PR #1912, opened 2025-06-15
Added addon Uptime kuma

bruvv:
You might see some teslamte stuff, that is what i am still working on but cannot get to work so i split my branches and switch to only uptime kuma since that is working
(and is very usefull addon to have!)


alexbelgium:
Thanks ! Ready to merge ? Did you check for data remanence upon restart ?

bruvv:
Ah did not check that let me verify :) give me a few minutes

bruvv:
@alexbelgium yes it retains data after an addon restart 👍 so ready for merge

bruvv:
the only thing I do not understand/can get working is the ingress. But that could be added in a future release

bruvv:
omg haha i should have read! It is already live: https://github.com/hassio-addons/addon-uptime-kuma i will close this PR.

alexbelgium:
ah well - for ingress you need to add a nginx proxy and map the ingress url to the internal url, I have several examples
But that would be for a future addon then ;-)

Good weekend

bruvv:
thanks for the tip :) will keep working on teslamate in the mean time

Have a good weekend too alex!

PR #1921, opened 2025-06-21
chore: upgrade vectorchord to 0.4.3 which is supported in latest immich version

danez:
In this version it is supported:

https://github.com/immich-app/immich/releases/tag/v1.135.0


alexbelgium:
Thanks ! Do we need to embed an upgrade script ?

PR #1926, opened 2025-06-24
Add healthcheckscript for postgres

danez:
Not sure if this is 100% correct but it is a follow up of #1925 

Please check if this is correct.


alexbelgium:
Thanks! And you taught me something, I didn't knew that bashio::config could set a default value if null

I think the last remaining thing is to be sure the script has +x permissions

danez:
Okay I changed the permissions.

alexbelgium:
Thanks! Now the last remaining element is to increment the versions in config.json in order to trigger the build

alexbelgium:
Thanks! let's try

Issue #1877, opened 2025-05-26
🐛 [Portainer] cant login / creat container / enter container etc.


Joriex:
### Description

The other day portainer just stopped working.
I cant login with correct username and password: error message:"Unable to Login"

Supervisor does not give any further error logs

My friends have the same problem, anyone else? I couldnt find any infos on that but its already going for about 4 days.

I already tried to reinstall portainer addon, same error

### Reproduction steps

```bash
1. Install portainer
2. try to login
```

### Addon Logs

```bash
Starting...
/etc/cont-init.d/00-banner.sh: executing
-----------------------------------------------------------
 Add-on: Portainer
 Manage your Docker environment with ease
-----------------------------------------------------------
 Add-on version: 2.30.1
 You are running the latest version of this add-on.
 System: Home Assistant OS 15.2
 Architecture: aarch64 / raspberrypi4-64
 Home Assistant Core: 2025.5.2
 Home Assistant Supervisor: 2025.05.3
-----------------------------------------------------------
 Please, share the above information when looking for help
 or support in, e.g., GitHub, forums
-----------------------------------------------------------
 Provided by: https://github.com/alexbelgium/hassio-addons 
-----------------------------------------------------------
/etc/cont-init.d/00-global_var.sh: executing
certfile='fullchain.pem'
keyfile='privkey.pem'
password=******
ssl='false'
/etc/cont-init.d/01-config_yaml.sh: executing
...
/etc/cont-init.d/01-custom_script.sh: executing
...
/etc/cont-init.d/30-nginx.sh: executing
/etc/cont-init.d/31-portainer.sh: executing
Starting the upstream container
s6-rc: info: service s6rc-oneshot-runner: starting
s6-rc: info: service s6rc-oneshot-runner successfully started
s6-rc: info: service base-addon-banner: starting
-----------------------------------------------------------
 Add-on: Portainer
 Manage your Docker environment with ease
-----------------------------------------------------------
 Add-on version: 2.30.1
 You are running the latest version of this add-on.
 System: Home Assistant OS 15.2  (aarch64 / raspberrypi4-64)
 Home Assistant Core: 2025.5.2
 Home Assistant Supervisor: 2025.05.3
-----------------------------------------------------------
 Please, share the above information when looking for help
 or support in, e.g., GitHub, forums or the Discord chat.
-----------------------------------------------------------
s6-rc: info: service base-addon-banner successfully started
s6-rc: info: service fix-attrs: starting
[...]

PR #1930, opened 2025-06-28
Fix Portainer connection upgrade

lmagyar:
Exec-ing into a container is still broken, we get the black terminal, but no prompt, because we have to copy Connection and Upgrade headers into the location section also, like the Host headers. I have no clue why... I haven't noticed this a few hours ago, I was happy with the black terminal "window", haven't noticed the missing prompt. :/

Additionally the 2nd location section for the websocket api is removed, it is unnecessary, it is identical with the first, I assume it was there, because originally it differed from the "/" location only by these upgrade headers, and yes, it is for the websockets, though doesn't harm in case of the whole site.

Related #1877


alexbelgium:
Thanks ! Let’s try

PR #1945, opened 2025-07-09
[Postgres] Add doc to make Postgres only available local

danez:
I wanted to make Postgres only available for other Add-ons. 

Maybe this is useful for others as well.


alexbelgium:
Thanks!

PR #2019, opened 2025-08-09
2018: adding option to clear codecs folder on addon start

### LINKED ISSUE #2018: ✨ [REQUEST] Plex
### Checked

- [x] If a new addon, I have checked on  Google that such as addon doesn't already exists
- [x] If a new feature, I have searched in the discussions & issues that it was not already covered

### Which addon?

hey @alexbelgium 

I had troubles with Plex and we came to the conclusion that it was the `codec` folder in `/share`.

The Trash Guides team suggested to clear that folder and afterwards it was good.

They also suggested me to use the base image of Hotio cause this one clears the `codec` folder on every restart.

I would see 2 options:
A) Your addon uses the hotio image as base (not sure how much else differs, so no idea about the amount of work needed).
B) Implement a codec folder wipe on each start of you addon, I guess that would be easy.

What do you think?

Best,
Dirk

Yes, described above

Described above

dm82m:
PR to fix #2018

### DIFF
--- a/plex/config.json
+++ b/plex/config.json
@@ -180,7 +180,7 @@
   "udev": true,
   "url": "https://github.com/alexbelgium/hassio-addons/tree/master/plex",
   "usb": true,
-  "version": "1.42.1.10054-f333bdaa8-ls277",
+  "version": "1.42.1.10054-f333bdaa8-ls277-2",
   "video": true,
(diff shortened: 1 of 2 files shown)


alexbelgium:
Thanks

PR #2078, opened 2025-08-29
Update immich to 1.140.0

danez:


alexbelgium:
Hi, it’s automatic update day so actually it is already applied ! But thanks for the pr

PR #2085, opened 2025-09-05
Add Mealie HA discovery message

andrew-codechimp:
This adds the required script to send a discovery message to Home Assistance allowing prompting for the Mealie Integration to be added, passing the IP/Port, easing user configuration.

The related PR for HA Core is [here](https://github.com/home-assistant/core/pull/151773), they can be released independently as the message is ignored if the integration does not support it yet.

### DIFF
--- a/mealie/CHANGELOG.md
+++ b/mealie/CHANGELOG.md
@@ -1,4 +1,4 @@
-## v3.1.3 (05-09-2025)
+## v3.2.2 (27-09-2025)
 - Send discovery message to Home Assistant
 
 ## v3.2.1 (20-09-2025)


alexbelgium:
Thanks ! Please let me know when it’s ready to merge

alexbelgium:
Don’t you need to add an api access in the config to allow setting the integration ?

andrew-codechimp:
Hi Alex, will do.  Just want to check the integration side is right in case I need to change things here.

No API access is required, this just passes the host/port to the integration, it will always be a manual process for the user to create an API token in Mealie then paste it into the discovery flow within the HA integration, but at least they don't have to struggle with working out the difference between IP/host name/ingress name.

I have tested this by using the mealie.sh code injector on my local environment.

This is a very similar approach to the official AdGuard addon which does discovery as well, not many examples about though.

andrew-codechimp:
Still awaiting review and merge of core integration PR

andrew-codechimp:
The mealie integration change has been merged into HA Core and will be included in the November release.
This change can be merged now, the message sent will just be ignored for now.
Not sure on the Action failures, I have rebased and updated the change log.

alexbelgium:
Thanks!

PR #2155, opened 2025-11-01
Update immich to 2.2.1

danez:
Also cleanup changelogs

The images do exist :) 
https://github.com/imagegenius/docker-immich/pkgs/container/immich


danez:
Hmmm did not work is still 2.1.0

Edit: Seems the aarch images are not updated, because upstream aarch images do not exists. Seems they have problems with the build: https://github.com/imagegenius/docker-immich/issues/635

PR #2223, opened 2025-11-22
Fix Scrutiny config example: Add required 'env_vars'

mausland:
Added clarification on env_vars

The current example causes a 'Missing option env_vars' error on save. Added env_vars: [] to fix validation.


alexbelgium:
Thanks !

PR #2398, opened 2026-01-24
Fix Ingress for changedetection.io

swiergot:


alexbelgium:
Thanks you also need to increment the version to trigger a build

swiergot:
@alexbelgium In config.yaml? That seems to be upstream version.

alexbelgium:
Yes indeed it’s the version value change that triggers the build you can set it to version: "0.52.9-2" for example

swiergot:
Hi, sorry for no response, been a bit busy, and thanks for doing this change.

Issue #2416, opened 2026-01-29
🐛 [Seafile] env_vars doesn't work


Lion62:
### Description

<img width="718" height="489" alt="Image" src="https://github.com/user-attachments/assets/f6903390-b873-44cd-b931-1bcb8c0a11c3" />

env_vars doesn't work on seafile

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Addon Logs

```bash
Seafile INFO: Setting variables CONFIG_LOCATION='/config/addons_config/seafile/config.yaml' FILE_SERVER_ROOT='http://homeassistant.local:8082' PGID='1000' PORT='8082' PUID='1000' SEAFILE_ADMIN_EMAIL='me@example.com' SEAFILE_ADMIN_PASSWORD='******' SERVER_IP='homeassistant.local' TZ='Europe/Paris' data_location='/media/seafile' database='mariadb_addon' env_vars='[ { "name": "DB_HOST", "value": "core-mariadb" }, { "name": "DB_USER", "value": "seafile" }, { "name": "DB_PASS", "value": "VOTRE_MOT_DE_PASSE_SQL" } ]' Error : /etc/cont-init.d/99-run.sh exiting 1 Starting the upstream container
```

### Architecture

amd64

### OS

HAos

alexbelgium:
Hi, this should be better

PR #2418, opened 2026-01-30
Fix Seafile env_vars export handling

alexbelgium:
### Motivation
- The add-on was exporting the entire `env_vars` option as a single environment variable which prevented individual custom variables from being applied correctly and caused the container init to fail.
- The intent is to apply each entry in the `env_vars` list as a proper environment variable available to the Seafile runtime and startup scripts.

### Description
- Skip the raw `env_vars` key when exporting generic add-on options from `/data/options.json` so it is not exported as a single JSON string. 
- Add a processing block that reads `.env_vars` from `options.json`, validates entries, iterates the list and exports each variable into the current environment with `export`.
- Escape values with `printf "%q"` and persist `export` lines into Seafile startup scripts under `/home/seafile/*.sh` and `*/opt/seafile/*` so downstream scripts see the variables at runtime.
- Document the fix in `seafile/CHANGELOG.md`.

### Testing
- No automated tests were executed for this change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_697cd2ab6dd883258317592e0c8fab17)

### DIFF
--- a/seafile/config.yaml
+++ b/seafile/config.yaml
@@ -128,5 +128,5 @@ services:
 slug: seafile
 udev: true
 url: https://github.com/alexbelgium/hassio-addons/tree/master/seafile
-version: "12.0.17"
+version: "12.0.18"
 webui: http://[HOST]:[PORT:8000]


alexbelgium:
https://github.com/alexbelgium/hassio-addons/issues/2416

PR #2432, opened 2026-02-02
[WHATSAPPER] fix 1.1.2

baldarn:
add 1.1.2 in config :/


alexbelgium:
Hi, looks like the config file is invalid couldn’t you order fix ? Also when you update the version in the config file you also need to update in the updater.json

Thanks very much !

baldarn:
the updater was already pointing the 1.1.2 from the previous pr

what is wrong in the config file?

sadly I don't have the plugin anymore. I use the services through docker compose :/
so I can't test this on my home assistant installation

alexbelgium:
Thanks I’ll do it

PR #2448, opened 2026-02-05
Implement logging for OpenVPN with configurable level

alexbelgium:
Add logging functionality with configurable log level for OpenVPN.

@litinoveweedle


litinoveweedle:
Very nicely done indeed! I was thinking about implementing logging of the wireguard logging as well, but unfortunately it doesn't seem to be feasible. Wireguard uses kernel logging and it doesn't seem to be accessible from within docker containers.

alexbelgium:
Thanks ! Would something like that help :

instead of _cmd "ip link set ${config["Interface"]} up" || return 1
```
if ! _cmd "ip link set ${config["Interface"]} up"; then
  [[ "$log_level" == "debug" ]] && bashio::log.warning "--- WireGuard debug (wg show) ---"
  [[ "$log_level" == "debug" ]] && wg show 2>&1 | while IFS= read -r l; do bashio::log.warning "$l"; done
  return 1
fi
```

PR #2463, opened 2026-02-08
[SignalK] Add I2C device mappings to SignalK AddOn configuration

eburi:
Add I2C device mappings to configuration. There are SignalK Plugins that can access i2c-devices, but without the device mapping the container does not have permission to open the devices.


alexbelgium:
Thanks

PR #2568, opened 2026-03-07
Deprecate Tor

akrigator:
Since a bridge support has been finally added in to the community version by [PR](https://github.com/hassio-addons/addon-tor/pull/170)


alexbelgium:
Thanks !

PR #2599, opened 2026-03-20
Adjusted TRANSMISSION_WEB_UI options in config

yannpub:
1. Added transmissionic webUI
2. moved transmission-web-control to the end - is removed in the current image version, but there are discussions ongoing so not removing it for the moment


alexbelgium:
Thanks ! If you want it reflected in your installed addon you need to increment the version number too

PR #2601, opened 2026-03-21
qBittorrent upnp and firewall for VPN

litinoveweedle:
Hello Alex.

This PR extends the VPN configuration we implemented together some time ago by adding UPnP for port mapping opening for qBittorrent. Implementation also includes simple firewall for incoming connections.

I recently started to use qBittorrent addon more and realized that torrent client port is not properly opened on the VPN service. In this case qBittorrent reports its Connection Status as 'firewalled'. Normally UPnP/NAT-PMP port openning/mapping is enabled by the UPnP switch in the qBittorrent configuration. Unfortunatelly qBittorrent native UPnP implementation has its limitations and very often does not work correctly with the VPN services (as it can't reliably detect VPN GW.

Screenshot from 2026-03-21 10-32-30

I therefore decided to extend VPN functinality by implementing UPnP port forwarding service (for VPNs which supports such functionality - like ProtonVPN). I based my work on the https://github.com/soxfor/qbittorrent-natmap project and implemented it as another S6 service (similar to vpn-monitor). I also added very simple firewall to VPN interface filtering incoming connections except for the configured qBittorrent torrent listening port. I also added configuration switch to manage this service.

This time I was able to test the work (at least for my wireguard VPN) and it works without issues. ;-) 

image

Some additional testing is suggested for openvpn based VPNs. Please note, that Connection Status will chnage only after first connection from the mapped port is established - so for testing you need to have some torent trafic.

### DIFF
--- a/qbittorrent/config.yaml
+++ b/qbittorrent/config.yaml
@@ -143,4 +143,4 @@ schema:
 slug: qbittorrent
 udev: true
 url: https://github.com/alexbelgium/hassio-addons
-version: "5.1.4-17"
+version: "5.1.4-18"


alexbelgium:
Thanks ! Let’s try

PR #2640, opened 2026-04-10
fix(birdnet-pipy): fix icecast crash and Live Feed broken in ingress mode

Suncuss:
## Summary

Two fixes for issues introduced by upstream BirdNET-PiPy features (file-based logging in v0.5.8, multi-source audio in v0.6.0) that only surface in the HA single-container environment.

### 1. Icecast crashes on startup (502 Bad Gateway)

`01-structure.sh` creates `/app/data/logs` as root, but icecast runs as `icecast2` via `gosu`. First log write fails → `set -e` kills the script → restart loop.

**Fix:** `chown icecast2:` the log dir/file in `services.d/icecast/run` before dropping privileges. Trailing colon picks the correct primary group (`icecast` on Debian, not `icecast2`).

### 2. Live Feed broken in ingress mode (404)

The `"/stream/"` sub_filter rule also matches `api.get("/stream/config")` in the JS bundle, double-prefixing the request: `/hassio_ingress/TOKEN/api/hassio_ingress/TOKEN/stream/config`.

**Fix:** Remove the `"/stream/"` sub_filter rules. Upstream (>= 0.6.2) now uses relative stream URLs that resolve via the existing `<base href>`.

### DIFF
--- a/birdnet-pipy/CHANGELOG.md
+++ b/birdnet-pipy/CHANGELOG.md
@@ -1,6 +1,8 @@
 
 ## 0.6.2 (2026-04-11)
 - Update to latest version from Suncuss/BirdNET-PiPy (changelog : https://github.com/Suncuss/BirdNET-PiPy/releases)
+- Fix Icecast crashing on startup due to log directory permissions (502 Bad Gateway on Live Feed)
+- Fix Live Feed broken in ingress mode — stream config request was double-prefixed by sub_filter
 
 ## 0.6.1 (2026-04-06)
 - Update to latest version from Suncuss/BirdNET-PiPy (changelog : https://github.com/Suncuss/BirdNET-PiPy/releases)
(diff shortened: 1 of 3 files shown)


alexbelgium:
Hi, thanks have you tested it to work ? Could you please also increment the version to launch the build

Suncuss:
yes, tested and working with latest 0.6.2 version. clarify on version bump, should I bump the version in config.yaml or you mean version bump on BirdNET-PiPy repo?

alexbelgium:
In the config.yaml as that triggers the addon build :) thanks !

Suncuss:
seems like a lot of checks for other addons are failing

alexbelgium:
Thanks ! Fails are due to residual things that don’t really matter I’ll fix it another time

alexbelgium:
You were pr 888 lucky number

PR #2644, opened 2026-04-13
Reorganise options in a more logical way

yannpub:
Reorganise options in a more logical way (config -> Username -> password)


alexbelgium:
Thanks - to have it appear in your system you must also increment the version in order to trigger an update

yannpub:
Thanks Alex !
There is nothing urgent in this change, it can wait the next major update, this is why I didn’t update the version to trigger immediate change.

alexbelgium:
Thanks !

PR #2645, opened 2026-04-13
bump manyfold version to 0.137.0

ToledoEM:
New Version of Manyfold with new fucntionalities, like comments and comments federation. 
_I dont care about those but why not_  :-) 

//
Enjoy


alexbelgium:
Hi, thanks! you could also tag latest so that you just need to increment the version in config, and it will build using the right on

ToledoEM:
No problem, I am getting my way arround docker... 👍

PR #2649, opened 2026-04-14
fix(birdnet-pipy): simplify ingress to single <base href> rewrite

Suncuss:
## Summary

- Aligns with upstream [Suncuss/BirdNET-PiPy v0.6.3](https://github.com/Suncuss/BirdNET-PiPy/releases/tag/v0.6.3), which declares `<base href="/">` in `index.html` and makes all internal URLs relative (Vite `base: './'`, axios `baseURL = BASE + 'api'`, socket.io `path = BASE + 'socket.io'`).
- Replaces seven `sub_filter` rules with one `<base href>` rewrite — removes the byte-match-in-JS-bundle brittleness that caused the `/stream/` double-prefix bug in 0.6.2.
- Version bumped to `0.6.3` to trigger CI.


alexbelgium:
Thanks ! Much better indeed

Suncuss:
Thanks for the ground work! I also cleaned up my timezone logic a bit in my codebase, so it does not depend on the TZ variable anymore. So i might work on anther PR for this addon to simplify that part as well.

alexbelgium:
Nice ! Thanks

Issue #2688, opened 2026-05-08
🐛 [photoprism] Access denied to MariaDB due to IPv6 local address after HAOS 17.3 installation


guandalf:
### Description

Hey Alex,

I am not sure this is an issue with the photoprism addon or with HAOS itself.

What happened is that, after the installation of HAOS 17.3, the photoprism app didn't come up with the following issue:

`ERROR 1045 (28000): Access denied for user 'service'@'fd0c:ac1e:2100::f' (using password: YES)
Error : /etc/cont-init.d/99-run.sh exiting 1`

and looking in MariaDB user table I see no users created that allow logging in from an IPv6 address (https://github.com/home-assistant/addons/blob/0eeb5ecfb4bee8d8f34813b891edd5d338494c8f/mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-post/run#L90)

What confuses me the most, it was working previously and I only see a change related to IPv6 in 17.2, but in fact IPv6 should be a thing since a lot (this would mean that the mariadb addon should be fixed to create users able to log in from IPv6 local addresses).

What do you think?

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Addon Logs

```bash
ERROR 1045 (28000): Access denied for user 'service'@'fd0c:ac1e:2100::f' (using password: YES)
Error : /etc/cont-init.d/99-run.sh exiting 1
```

### Architecture

aarch64

### OS

HAos

guandalf:
Now the app starts correctly. I have briefly reviewed the MR, but I am not skilled enough in HA coding to tell if the resolution is the best approach.

alexbelgium:
Hi, I didn’t have much time and used an AI copilot (Claude opus 4.7) but glad it works !

PR #2693, opened 2026-05-10
fix: force IPv4 host resolution for MariaDB addon connections (#2688)

alexbelgium:
On HAOS >=17.3 the Supervisor Docker network gained IPv6, so
core-mariadb resolves to an IPv6 address first. The official MariaDB
addon only grants its service user from the IPv4 supervisor subnet, so
connections from IPv6 fail with "Access denied".

Resolve the hostname to its IPv4 address before connecting in every
addon that consumes bashio::services 'mysql' 'host': photoprism,
monica, fireflyiii, seafile, zoneminder. Fall back to the raw hostname
if resolution fails so IPv4-only setups keep working unchanged.

### DIFF
--- a/fireflyiii/CHANGELOG.md
+++ b/fireflyiii/CHANGELOG.md
@@ -1,4 +1,7 @@
 
+## 6.6.2-1 (2026-05-10)
+- Fix MariaDB connection on HAOS >=17.3 by forcing IPv4 host resolution (#2688)
+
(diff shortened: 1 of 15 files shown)


alexbelgium:
@copilot apply changes based on the comments in [this thread](https://github.com/alexbelgium/hassio-addons/pull/2693#pullrequestreview-4258838852)

PR #2716, opened 2026-05-17
Manyfold version bump

ToledoEM:
- bump version to 0.140.1
- better readme example for configuration in small systems. In my case RPi5 2 GB ram


alexbelgium:
Thanks

PR #2725, opened 2026-05-21
aurral: v1.76.17-1 - fix config schema, add weekly_flow_folder option, clean up build.json

petruknw:
Follow-up to #2723 addressing the issues flagged after merge.

## Changes

- `env_vars` → `environment` (correct HA addon config schema)
- Remove `squash: false` from `build.json` (default value)
- Remove redundant `labels` block from `build.json`
- Add `image: ghcr.io/alexbelgium/aurral-{arch}` to `config.yaml` to trigger build pipeline
- Add `weekly_flow_folder` option (configurable subfolder suffix appended to `download_folder` for weekly flow files, default: `weekly-flow`)
- Fix data/downloads paths to match upstream volume layout (`/app/backend/data`, `/app/downloads`)
- Version bump to `1.76.17-1`
- README and CHANGELOG updated

PR #2726, opened 2026-05-21
aurral: v1.76.17-2 - fix entrypoint, restore mkdirs, clean up build.json

petruknw:
Follow-up to #2725.

## Changes

- Fix entrypoint: replace `bashio::config` (not available in upstream node image) with HA-injected env vars
- Restore `mkdir` for persistent data and download directories
- Remove redundant `labels` block from `build.json`
- Version bump to `1.76.17-2`

### DIFF
--- a/aurral/CHANGELOG.md
+++ b/aurral/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.76.17-2
+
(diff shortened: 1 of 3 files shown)


petruknw:
Sorry alexbelgium, the previous PR broke startup.

PR #2751, opened 2026-06-01
aurral: fix CMD path server.js -> backend/server.js

petruknw:
## Problem

As of v1.76.42, aurral fails to start with:
```
Error: Cannot find module '/app/server.js'
```

## Root cause

Upstream restructured the project — `server.js` moved from `/app/server.js` to `/app/backend/server.js` (see [upstream Dockerfile](https://github.com/lklynet/aurral/blob/main/Dockerfile)). Our addon's `Dockerfile` has a hardcoded `CMD ["node", "server.js"]` that overrides the upstream entrypoint and now points to the wrong path.

## Fix

Update `CMD` to `["node", "backend/server.js"]` to match the upstream path.


alexbelgium:
Thank

PR #2773, opened 2026-06-16
Fix wger database location after update to the latest version

rojuvi:
Update settings file lookup filter since settings.py was moved to several python files within settings folder


alexbelgium:
Thanks ! Did you test it, and can you please increment the changelog and version in config.yaml in order to launch a build

rojuvi:
> Thanks ! Did you test it, and can you please increment the changelog and version in config.yaml in order to launch a build

Hey! I tested the command directly on the container terminal through portainer.
I will update the changelog and version, thanks

PR #2789, opened 2026-06-20
allow using seerr without admin

sanjay900:
I have some family on my home assistant instance that I'd like to give access to seerr, but right now it's missing the panel admin flag, so they can't see it without admin.

Notes that radarr and sonarr already have this flag, so it seems reasonable for it to be here too.


alexbelgium:
Thanks

PR #2832, opened 2026-07-09
Update Bazarr, Prowlarr and qBittorrent addons

juhenriquez:
## Summary
- Update Bazarr from `1.5.6` to `1.6.0`.
- Update Prowlarr from `nightly-2.5.0.5453-ls4` to `nightly-2.5.1.5460-ls4`.
- Update qBittorrent from `5.2.2` to `5.2.3`.
- Refresh `updater.json` metadata, changelog entries, and `BUILD_UPSTREAM` where present.

## Why
These add-ons from the requested list were behind their tracked upstream releases.

## Validation
- Ran `git diff --check`.
- Parsed the updated `updater.json` files successfully.
- Verified upstream releases:
  - https://github.com/linuxserver/docker-bazarr/releases/tag/v1.6.0-ls354
  - https://github.com/linuxserver/docker-prowlarr/releases/tag/nightly-2.5.1.5460-ls4
  - https://github.com/linuxserver/docker-qbittorrent/releases/tag/5.2.3_v2.0.13-ls467

## Summary by CodeRabbit

* **Chores / Updates**
  * Updated Bazarr, Prowlarr, and qBittorrent to newer upstream releases, including refreshed version/date metadata so the latest builds show correctly in the catalog.
  * Synced each add-on’s packaged release information with the latest upstream container/image versions.

* **Documentation**
  * Added new top changelog entries for Bazarr (1.6.0), Prowlarr (nightly 2.5.1.5460-ls4), and qBittorrent (5.2.3), each linking to the corresponding upstream release notes.

PR #2833, opened 2026-07-09
Fix builder after app updates

juhenriquez:
﻿# Proposed Changes

Fix the builder failures seen after #2832 was merged:

- remove `webui` from Bazarr and Prowlarr because Ingress is enabled and the Home Assistant add-on linter rejects that combination
- download VueTorrent from the stable `releases/latest/download/vuetorrent.zip` URL instead of scraping the GitHub releases API, which returned an empty URL during the aarch64 qBittorrent build

These fixes should allow the Builder workflow to publish the new Bazarr, Prowlarr, and qBittorrent images that Home Assistant is trying to pull.

## Related Issues

Follow-up to #2832.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed explicit web interface URL templates from two add-on configurations, relying on the add-on’s default web UI behavior instead.
  * Improved the torrent client’s bundled VueTorrent web UI download by switching to a direct “latest release” link, improving reliability during setup/initialization.

### DIFF
--- a/bazarr/config.yaml
+++ b/bazarr/config.yaml
@@ -113,4 +113,3 @@ slug: bazarr_nas
 udev: true
 url: https://github.com/alexbelgium/hassio-addons/tree/master/bazarr
 version: "1.6.0"
-webui: "[PROTO:ssl]://[HOST]:[PORT:6767]"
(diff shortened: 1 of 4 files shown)


alexbelgium:
Thanks !

PR #2865, opened 2026-07-15
fix(claude_desktop): headroom wrapper path + TokenSave startup corruption hardening

alexbelgium:
## Summary
- `/usr/local/bin/claude` wrapper hardcoded `HEADROOM_BIN=/usr/local/bin/headroom`; actual binary lives at `/usr/bin/headroom`, so the `-x` check always failed and terminal Claude Code sessions never routed through the Headroom proxy. Resolved via `command -v headroom`.
- Startup TokenSave indexing chose init-vs-sync purely on file existence, so an interrupted `init` or a hard add-on stop could leave a corrupt graph that failed every subsequent boot. Prepare each repo under a startup-scoped `flock`, retry incremental `sync` (transient `SQLITE_BUSY` != corruption), and quarantine+rebuild only a genuinely unreadable or half-written (sentinel-flagged) index.

Bumps `claude_desktop` to 1.24 with matching CHANGELOG entries.

## Test plan
- [x] Wrapper fix verified byte-identical to the already-patched reference `/usr/local/bin/claude` in the running container
- [x] `shellcheck` clean on both scripts (one intentional SC2016 annotated — single-quoted heredoc by design)
- [x] `bash -n` syntax check on outer script and extracted inner abc routine
- [x] TokenSave routine runtime-tested against the real `tokensave` binary across 5 cases: fresh repo (init), healthy repo (sync, no quarantine), corrupted DB (retry → quarantine → rebuild), interrupted init via sentinel (rebuild), concurrent run under held lock (clean skip)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed terminal Claude Code sessions not routing through Headroom when the executable is installed in a nonstandard location.
  - Improved TokenSave startup recovery to prevent interrupted or corrupted indexing from causing persistent failures.
  - Added safer retries and automatic rebuilding of unreadable indexes.
- **Chores**
  - Updated the add-on configuration and changelog for version 1.24.

### DIFF
--- a/claude_desktop/CHANGELOG.md
+++ b/claude_desktop/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.24 (15-07-2026)
+
+- Fix the `/usr/local/bin/claude` wrapper never routing terminal Claude Code sessions through the Headroom proxy: it hardcoded `HEADROOM_BIN="/usr/local/bin/headroom"` while the binary is installed at `/usr/bin/headroom`, so the executable check always failed and the wrapper fell back to launching Claude Code directly. Resolve the binary with `command -v headroom` instead.
+
 ## 1.23 (15-07-2026)
 
 - Add a `ha-cli` helper that lets Claude configure Home Assistant (automations, scripts, scenes, helpers, dashboards, area/label/floor/entity registries, and service calls) through the Home Assistant Core API instead of a filesystem mount. It authenticates automatically with the add-on's `SUPERVISOR_TOKEN` via the Supervisor Core-API proxy (no token setup), and deliberately cannot reach `configuration.yaml`/`secrets.yaml` or other add-ons' credentials. Toggle with the new `enable_ha_api_helper` option (default on), which also controls a managed guidance block appended to `~/.claude/CLAUDE.md`.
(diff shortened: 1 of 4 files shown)


alexbelgium:
Addressed both findings in a follow-up commit on #2869 (that branch carries this code forward from master, so it's the live copy): b8c7cc381

- **flock -n silently skipping recovery:** switched to `flock -w 60` — waits up to 60s for the holder rather than skipping outright. Bounded rather than an unconditional blocking wait (as suggested) because a held lock only clears when its owner exits or dies (kernel-released), so a genuinely stuck lock is the one case an unbounded wait can't recover from, and cont-init scripts shouldn't be able to hang add-on boot indefinitely.
- **Quarantine on any failure:** now captures sync's stderr and only quarantines when it names actual database corruption (SQLite's "malformed"/"not a database"/"disk image" wording). Any other failure (permissions, disk full, transient SQLITE_BUSY from an external writer — the exact scenario @chatgpt-codex-connector flagged separately on this PR) leaves the index untouched and retries next start.

Since this PR is already merged, I didn't push here — see #2869 for the actual diff.

Issue #869, opened 2023-06-12
🐛 [EPIC GAMES FREE] Captcha not working


ajcorrea:
### Description

Just installed and configured addon:
config.json
`
{
    "runOnStartup": true,
    "cronSchedule": "15 18 * * *",
    "logLevel": "info",
    "webPortalConfig": {
        "baseUrl": "http://192.168.0.114:3005",
    },
    "accounts": [
        {
            "email": "my-epic-games-login@gmail.com",
            "password": "my-password"
        },
    ],
    "notifiers": [
    {
        "type": "email", // This indicates the notifier type
        "smtpHost": "smtp.gmail.com",
        "smtpPort": 587,
        "emailSenderAddress": "my-email-gmail-account@gmail.com",
        "emailSenderName": "[HomeAssistant]: Epic Games Captchas",
        "emailRecipientAddress": "my-other-email@gmail.com",
        "secure": true,
        "auth": {
            "user": "my-email-gmail-account@gmail.com",
            "pass": "my-password",
        },
    },
    ],
}
`

and gettings this error:

`
Starting...
/etc/cont-init.d/00-banner.sh: executing

-----------------------------------------------------------
 Add-on: Epic Games Free
 Automatically login and redeem promotional free games from Epic Games Store
-----------------------------------------------------------
 Add-on version: bullseye-slim-2023-05-03
 You are running the latest version of this add-on.
 System: Debian GNU/Linux 11 (bullseye)  (amd64 / qemux86-64)
 Home Assistant Core: 2023.6.0
 Home Assistant Supervisor: 2023.06.1
-----------------------------------------------------------
 Please, share the above information when looking for help
 or support in, e.g., GitHub, forums
-----------------------------------------------------------
 Provided by: https://github.com/alexbelgium/hassio-addons 
-----------------------------------------------------------
/etc/cont-init.d/01-custom_script.sh: executing
[17:02:02] INFO: Execute /config/addons_autoscripts/epicgamesfree.sh if existing
[17:02:02] INFO: ... no script found
/etc/cont-init.d/20-folders.sh: executing
/etc/cont-init.d/99-run.sh: executing
[17:02:02] WARNING: The config.json file found in /config/addons_config/epicgamesfree will be used. Please customize according to https://github.com/claabs/epicgames-freegames-node#json-configuration and restart the add-on
 
[17:02:02] INFO: Starting the app
 

> epicgames-freegames-node@4.1.0 entrypoint-config
[...]

PR #2869, opened 2026-07-15
fix(claude_desktop): Headroom zero savings — cowork session routing + Kompress activation

alexbelgium:
## Summary
Headroom kept reporting **zero savings** even after the wrapper path fix (#2865). Two independent root causes, both fixed:

1. **Desktop cowork/local-agent-mode sessions never reach the proxy.** Desktop spawns its bundled Claude Code binary at an absolute path (bypassing the `/usr/local/bin/claude` PATH wrapper) with `ANTHROPIC_BASE_URL` pinned to the production endpoint (headroom #869). The add-on now manages `env.ANTHROPIC_BASE_URL = http://127.0.0.1:8787` in `~/.claude/settings.json` — Claude Code writes settings `env` entries over the inherited environment at startup, and cowork sessions run with `--setting-sources=user,project,local`. Managed-value semantics: never clobbers a user-customized endpoint; removed when headroom or `headroom_wrap_claude_code` is disabled.

2. **Even proxied traffic compressed 0 tokens** (observed: 175 requests, 0 saved). The proxy's startup preload is deliberately cache-only, but the HF model cache defaulted to `~/.cache` → tmpfs, wiped every restart — so the Kompress ONNX model (`chopratejas/kompress-v2-base`) and its separately fetched `answerdotai/ModernBERT-base` tokenizer were never cached and the engine idled in "deferred" mode forever (misleadingly logged as `Kompress: not installed`). `svc-headroom` now sets `HF_HOME` to persistent `~/.headroom/hf` (~270 MB once) and pre-warms the cache at startup, bounded at 300 s so an offline install still starts the proxy in pass-through mode and retries next boot. The `proxy` extra's ONNX runtime is sufficient — the multi-gigabyte PyTorch `[ml]` extra is deliberately **not** added to the image.

Bumps `claude_desktop` to 1.27 with a CHANGELOG entry.

## Test plan (all verified live in a running add-on container)
- [x] settings.json env merge logic tested across 8 cases: fresh file, idempotent re-run (no rewrite), hooks preserved, user-custom URL left untouched (add + remove), managed URL removed cleanly, corrupt JSON → `.bak` + rebuild, unrelated env keys preserved
- [x] Model + tokenizer pre-warmed into persistent `HF_HOME`; proxy restart logs `Kompress: ENABLED (ModernBERT token compressor)` (previously `Kompress: not installed`)
- [x] Cache-only load path (`_load_kompress(allow_download=False)`) succeeds after pre-warm — exact code path the startup preload uses
- [x] End-to-end: `claude -p` round-trip through the proxy increments `/stats` `api_requests` 0 → 1
- [x] `bash -n` + shellcheck clean on both edited scripts

## Summary by CodeRabbit

* **New Features**
  * Claude Code sessions can be routed through the Headroom proxy when enabled.
  * Persistent Hugging Face model caching is now configured for Headroom to improve Kompress behavior.
  * Settings updates are idempotent and won’t overwrite existing, user-defined Claude Code endpoint configuration.
* **Bug Fixes**
  * Improved startup resilience for TokenSave per-repository index sync, quarantining/rebuilding only when corruption is detected (with safer concurrent restart handling).
* **Maintenance**
  * Updated the desktop add-on version to **1.27** and clarified installed package set expectations.


alexbelgium:
Fixed both actionable findings, thanks — and skipping the third with a reason:

**Codex (proxy health fallback, defeated by pinned env):** confirmed and fixed in b3c27024d. Removed the blocking Kompress pre-warm from svc-headroom/run entirely — the proxy's own request path (content_router.py) already calls ensure_background_load() on a cold cache and passes the request through uncompressed until the model lands, so the port binds immediately either way. That was the real fix; the pre-warm was unnecessary complexity that also introduced the regression you caught.

**CodeRabbit (Path.home() / ${HOME} resolve to /root):** skipping — this is a false positive for this codebase specifically. cont-init.d and s6-rc.d scripts here run with `#!/usr/bin/with-contenv bashio`, and 20-folders.sh (numbered before 82-claude_tools.sh, so it always runs first) writes HOME into the S6 container environment (`/run/s6/container_environment/HOME`) that with-contenv exports, in addition to sed-inserting an explicit `export HOME=$LOCATION` at the top of every s6-rc.d/*/run file. Python's `Path.home()` (via os.path.expanduser) honors $HOME from the environment before ever consulting the passwd database, regardless of effective UID. This is empirically proven in production: the pre-existing rtk-hook block in this same file (unmodified by this PR) uses the identical `Path.home()` pattern and correctly populates ~abc/.claude/settings.json today. The suggested fix (hardcode `/data/data`) would actually be a regression — it ignores the add-on's `data_location` option (users can point storage at /share, /config, or /mnt), silently writing settings outside a user's chosen persistent path for anyone who customizes it.

Also pushed a separate commit addressing the two TokenSave findings left over from #2865 (that code is inherited into this branch) — bounded `flock -w 60` instead of non-blocking skip, and quarantine now gated on an actual corruption signature in sync's stderr rather than any 3x failure.

PR #2872, opened 2026-07-16
fix(claude_desktop): proxy Selkies API websocket in ingress

alexbelgium:
## Summary

- Replace the inherited Selkies nginx ingress rewrite with an explicit Home Assistant ingress configuration.
- Proxy current Selkies WebSocket/API traffic at `/api/` to the local Selkies server on port 8082.
- Keep `/websocket` proxied to 8082 for compatibility with older Selkies/noVNC clients.
- Declare `ingress_port: 3001` explicitly, bump the add-on to `1.31`, and document the fix in `CHANGELOG.md`.

## Root cause

The add-on reaches `Data WebSocket Server listening on port 8082`, so the Selkies backend is running. The browser nevertheless never appears as connected and remains on **Waiting for stream**.

Current Selkies WebSocket mode opens the browser data socket under `<base>/api/websockets`, while the existing ingress script inherited the older linuxserver template, which only proxies `/websocket`. The dashboard HTML therefore loads through Home Assistant ingress, but its current data WebSocket never reaches Selkies.

The `/dev/dri/card0` EGL warnings are not the cause of this symptom: Selkies successfully initializes with the software `x264enc` encoder and starts its data server. The missing browser WebSocket connection is the blocking failure shown by the log.

## Validation

- `bash -n` passes for the rewritten `90-ingress.sh`.
- The generated nginx server block passes `nginx -t` with nginx 1.26.3.
- Compared against `master`: branch is ahead by four commits and changes only:
  - `claude_desktop/CHANGELOG.md`
  - `claude_desktop/config.yaml`
  - `claude_desktop/rootfs/etc/cont-init.d/90-ingress.sh`
- Diff statistics: changelog +4 lines; config +2/-1; ingress script +90/-19.

Runtime validation still requires rebuilding/restarting add-on version `1.31` and opening it through Home Assistant ingress.


alexbelgium:
Closing in favor of #2873, which investigates the same "waiting for stream" hang from a different angle.

I checked this PR's stated root cause against the actual pinned Selkies client source (`selkies-project/selkies` commit `0d134b6e1ffe42a579bc66363b0e7159ab22aacc`, which is what `docker-baseimage-selkies` builds into the image). `selkies-core.js` → `selkies-ws-core.js` builds the WebSocket URL purely from `window.location.pathname` with `'websockets'` appended — there's no hardcoded `/api/` prefix anywhere in the client bundle:

```js
const pathname = window.location.pathname.substring(0, window.location.pathname.lastIndexOf('/') + 1);
let websocketEndpointURL = new URL(`${ws_protocol}${window.location.host}${pathname}`);
websocketEndpointURL.pathname += 'websockets';
```

Separately, even if the browser did request something under `/api/hassio_ingress/<token>/websockets`, Home Assistant's Supervisor ingress proxy strips the `/api/hassio_ingress/<token>` prefix before forwarding to the add-on's internal nginx — that's why `90-ingress.sh` (in both the old template and this PR's rewrite) hardcodes `SUBFOLDER` to `/` rather than the real ingress path. So the add-on's own nginx only ever sees `/websockets`, which the existing `location /websocket` block already prefix-matches. The new `/api/` proxy location this PR adds shouldn't be reachable in practice.

#2873 traces the regression through the add-on's own git history instead (the desktop/nginx/GPU files are byte-for-byte unchanged since the last known-working version, 1.27 — only Claude-tooling scripts changed) and finds a real, confirmed bug: a service-startup race in the base image's s6 dependency graph where the graphical services can start before `init-video` grants GPU access, matching the `libEGL … Permission denied` in the boot log. It also pins the previously-rolling base image tag so the runtime stops drifting under future rebuilds.

Full writeup in #2873.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2887, opened 2026-07-21
Fix PR Check Build workflow: github.event.before is empty on pull_request events

alexbelgium:
## Bug

`check-addon-changes` (in `.github/workflows/onpr_check-pr.yaml`) computes changed addons by diffing against `${{ github.event.before }}`. That field is only populated on `push` events — this workflow triggers exclusively on `pull_request` (see the `on:` block), where `github.event.before` is always empty.

The result, every time:

```
git fetch origin "" || true
git diff --name-only "" "<sha>"
fatal: ambiguous argument '': unknown revision or path not in the working tree.
```

`changed_config_files` / `changed_changelog_files` come back empty, `changed_addons` resolves to `[]`, and the three downstream jobs (`check-changed-changelog`, `addon-linter`, `check-build`) are all gated on `changedAddons != '[]'` — so they show as SKIPPED regardless of what a PR actually touches. Confirmed on PR #2886 (run 29856484887) and cross-checked against #2843 — same skip pattern on both, so it's broken for every `pull_request`-triggered run, not one branch.

## Fix

Replace `github.event.before` with `github.event.pull_request.base.sha` in all 5 occurrences (2 in `find_addons`, 3 in `changed-files`) — that field is always populated for `pull_request` events and points at the correct base commit to diff against.

Since the `on:` trigger is `pull_request`-only (no `push`), this is a straight substitution — no conditional needed for a shared push/PR trigger.

## Checkout depth

Checkout steps have no `fetch-depth` set (default: 1, shallow). I left this alone rather than adding `fetch-depth: 0`, because:
- `git diff A B` only needs both commits' tree objects locally, not full ancestry — it doesn't require a non-shallow history.
- `git fetch origin <sha>` (already the pattern used here) fetches an arbitrary reachable commit directly; GitHub.com supports this (`uploadpack.allowReachableSHA1InWant`), so the base commit lands locally as a new shallow point even though the clone stays shallow depth-1 otherwise.
- Setting `fetch-depth: 0` on all three checkout steps (this job's, `addon-linter`'s, `check-build`'s matrix builds) would pull full repo history on every PR run in a 120+ addon monorepo for no benefit — the other two checkouts never diff, they just need the working tree.

Confirmed this holds in practice below rather than assuming it.

## Verification

**1. Local reproduction of the exact script logic**, using a real historical base/head SHA pair (`c2e61760` → `57366eea`, a commit that changed both `webtop_kde/config.yaml` and `webtop_kde/CHANGELOG.md`):

```
$ git fetch origin c2e6176088087217d41a8ed1dd9f7e671a870684
 * branch  c2e6176088087217d41a8ed1dd9f7e671a870684 -> FETCH_HEAD
$ git diff --name-only c2e61760... 57366eea... | grep -E '^[^/]+/config\.(json|ya?ml)$'
webtop_kde/config.yaml
Changed addons: ["webtop_kde"]
$ git diff --name-only c2e61760... 57366eea... | grep -iE '^([^/]+/)?changelog\.(md|txt|ya?ml|json)$'
webtop_kde/CHANGELOG.md
```

Confirms: with a populated base SHA, fetch-by-SHA succeeds and the diff/grep/detection pipeline correctly identifies the changed addon.

**2. This PR's own live run (confirmed)** — since this workflow is `pull_request`-triggered, opening this PR ran `check-addon-changes` with the *fixed* workflow file. Actual log output from that run:

```
Run git fetch origin "e6f6c95115cee777f768b2d88c184df15ed6d3ff" || true
From https://github.com/alexbelgium/hassio-addons
 * branch  e6f6c95115cee777f768b2d88c184df15ed6d3ff -> FETCH_HEAD
Changed config files:

Changed addons: []
```

No `fatal: ambiguous argument` — the fetch-by-SHA-into-a-shallow-clone approach works as expected in the real runner environment, confirming `fetch-depth: 0` isn't needed. `check-addon-changes` completed with `conclusion: success`; the three downstream jobs correctly show `skipped` for *this* PR (it only touches the workflow file itself, not any `*/config.*` or changelog — so an empty `changed_addons` here is the right answer, not the bug). The bug was the crash, not this particular empty result.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pull request change detection by deriving the comparison baseline from the pull request’s merge-base target revision.
  * Updated changelog and configuration change discovery to reliably compute and export consistent file lists for downstream validation jobs.
  * Ensured the workflow has sufficient history available to perform the corrected diff-based checks.

### DIFF
--- a/.github/workflows/onpr_check-pr.yaml
+++ b/.github/workflows/onpr_check-pr.yaml
@@ -22,8 +22,12 @@ jobs:
       - name: Find changed addon directories
         id: find_addons
         run: |
-          git fetch origin "${{ github.event.pull_request.base.sha }}"
-          diff_files=$(git diff --name-only "${{ github.event.pull_request.base.sha }}" "${{ github.sha }}")
+          # github.event.pull_request.base.sha is a snapshot from event-trigger time and can be
+          # stale if master advances before checkout; HEAD^1 is the actual base this merge
+          # commit was built against, so it's always correct.
+          base_sha=$(git rev-parse HEAD^1)
+          git fetch origin "$base_sha"
+          diff_files=$(git diff --name-only "$base_sha" "${{ github.sha }}")
           changed_config_files=$(printf '%s\n' "$diff_files" | grep -E '^[^/]+/config\.(json|ya?ml)$' || true)
           echo "Changed config files:"
           echo "$changed_config_files"
@@ -33,8 +37,9 @@ jobs:
       - name: Find changelog
         id: changed-files
         run: |
-          git fetch origin "${{ github.event.pull_request.base.sha }}"
-          diff_files=$(git diff --name-only "${{ github.event.pull_request.base.sha }}" "${{ github.sha }}")
+          base_sha=$(git rev-parse HEAD^1)
+          git fetch origin "$base_sha"
+          diff_files=$(git diff --name-only "$base_sha" "${{ github.sha }}")
           changed_changelog_files=$(printf '%s\n' "$diff_files" | grep -iE '^([^/]+/)?changelog\.(md|txt|ya?ml|json)$' || true)
           echo "$changed_changelog_files"
           {
--- a/.github/workflows/onpr_check-pr.yaml
+++ b/.github/workflows/onpr_check-pr.yaml
@@ -19,6 +19,10 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v7
+        with:
+          # Need the merge commit's parents resolvable (HEAD^1 below): a depth-1 shallow
+          # checkout truncates parent refs entirely at the boundary commit.
+          fetch-depth: 2
       - name: Find changed addon directories
         id: find_addons
         run: |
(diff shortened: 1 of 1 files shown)


alexbelgium:
Addressed in f6a9ac2:

- **coderabbitai / codex — masked fetch/diff failures**: `git fetch`/`git diff` no longer have a blanket `|| true`. The diff is now captured into a variable first, then filtered with `grep ... || true` separately — so grep's expected "no match" exit code is still tolerated, but a real fetch or diff failure now aborts the job instead of silently producing an empty `changed_addons`.
- **coderabbitai / codex — multiline changelogs_files output**: now written with the `GITHUB_OUTPUT` multiline delimiter syntax instead of a plain `key=value` echo, so a PR touching more than one addon's `CHANGELOG.md` no longer corrupts the output file.

Verified locally: a crafted 3-line value round-trips correctly through the new delimiter block; a real 30-commit range spanning 10 changelogs across 18 addons produces the correct multiline output and `changed_addons` list; a fetch against a nonexistent SHA now aborts the script (exit 128) instead of being swallowed.

PR #2888, opened 2026-07-22
fix(ci): scope revert-on-failure to the failing push's own commits

alexbelgium:
## Summary

The `revert-on-failure` job in `onpush_builder.yaml` was reverting more commits than the failing push actually introduced.

It re-fetched and checked out `master`, then computed the revert range as `before..HEAD` — but by that point `HEAD` was the *live* tip of master, not the head of the failing push. The weekly `addons_updater` bot pushes each addon as its own commit, seconds apart, and each push triggers its own Builder run. When one addon's build failed, the revert job for that push often didn't finish until several *later, successful* addon-bump commits had already landed on master — so `before..HEAD` spanned all of them, and they all got reverted along with the actually-failing one.

Example from `master` history: `ente` failed to build at 05:27:29 UTC. By the time its `revert-on-failure` job ran (finishing 05:29:51), 13 more addons had successfully built and merged. All 14 got reverted in one shot — only `prowlarr`, which landed after the revert push, survived.

## Fix

- Revert `before..github.sha` (the commits this specific push introduced) instead of `before..HEAD`.
- Retry the final push with a rebase loop, since master can still move again between reverting and pushing (e.g. another concurrent updater-bot push).

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/onpush_builder.yaml'))"` — YAML parses cleanly
- [ ] Next time the weekly updater runs, confirm a single addon build failure only reverts that addon's commit, not neighboring successful ones

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic recovery when a build fails by accurately reverting only the commits introduced by the failed push.
  * Added safer retry and synchronization logic to keep recovery commits consistent when updates occur concurrently.
  * Recovery now fails fast with a clear error message if it cannot complete after multiple push attempts.

### DIFF
--- a/.github/workflows/onpush_builder.yaml
+++ b/.github/workflows/onpush_builder.yaml
@@ -55,6 +55,9 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+    outputs:
+      sanitizeCommitted: ${{ steps.sanitize_commit.outputs.committed }}
+      sanitizeCommitSha: ${{ steps.sanitize_commit.outputs.commit_long_sha }}
     steps:
       - uses: actions/checkout@v7
         with:
@@ -90,6 +93,7 @@ jobs:
         uses: ymwymw/check-mixed-line-endings@v2
 
       - name: Commit sanitize changes
+        id: sanitize_commit
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
         uses: EndBug/add-and-commit@v10
         with:
@@ -410,6 +414,8 @@ jobs:
         env:
           BEFORE: ${{ github.event.before }}
           HEAD_SHA: ${{ github.sha }}
+          SANITIZE_COMMITTED: ${{ needs.prebuild-sanitize.outputs.sanitizeCommitted }}
+          SANITIZE_SHA: ${{ needs.prebuild-sanitize.outputs.sanitizeCommitSha }}
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


alexbelgium:
Thanks for the review — here's the disposition of each finding, in 7d7ece7:

**Fixed:**
- Rebase conflict during the push-retry loop was previously swallowed by `set -e`, silently burning the remaining retry attempts. Now aborts the rebase and fails loudly on a real conflict.
- `prebuild-sanitize`'s own `[nobuild]` commit lands on top of `HEAD_SHA` within the same run, so `before..HEAD_SHA` didn't include it. It's still this push's own fallout, not a neighboring push's, so its SHA is now captured via job output (`commit_long_sha`) and reverted explicitly alongside the original commit. This keeps the revert scoped to "this push" rather than widening the range back to a moving target — which is the exact bug this PR fixes, since the updater bot fires many pushes in quick succession and a moving-target range sweeps in unrelated, successful ones.

**Left alone (pre-existing, not introduced by this PR):**
- All-zero `before` on a newly-created `master` push: this fallback is unchanged from the original code, and `master` has existed for years, so it's practically unreachable.
- Merge commits need `git revert -m`: confirmed this is a real, active gap (`master` does get genuine 2-parent merge commits from normal PR merges), but it's identical in the code today, before this PR touched anything. Filing a separate follow-up rather than folding an unrelated fix in here.

PR #2908, opened 2026-07-27
fix(ci): pass github_token to claude-code-action so AI workflows can auth

alexbelgium:
## Problem

[Run 30226365228](https://github.com/alexbelgium/hassio-addons/actions/runs/30226365228/job/89857101280) reported success but did nothing. The `Classify` step is `continue-on-error: true`, so the 401 inside it was swallowed and the job went green:

```
Auto-detected mode: agent for event: issues
Requesting OIDC token...
OIDC token successfully obtained
Exchanging OIDC token for app token...
App token exchange failed: 401 Unauthorized - User does not have write access on this repository
[x3 attempts]
##[error]Action failed with error: User does not have write access on this repository
```

followed by `::warning::no usable verdict produced, leaving issue untouched`. That is why triage looked idle rather than failing.

## Cause

Every `claude-code-action` step except the one in `on_claude_mention.yml` left the **`github_token` input** unset. These workflows do set `GH_TOKEN` as a *step env* — that is what the CLI's `gh` calls use, but it is not the action input.

With the input unset the action falls back to the OIDC → Claude App token exchange, and that exchange validates that **`github.actor` has write access on the repository**. On an `issues.opened` event `github.actor` is the outside reporter who filed the issue, so it can never pass.

Confirmed against the pinned SHA `44423bd`:

- `action.yml:287` — `OVERRIDE_GITHUB_TOKEN: ${{ inputs.github_token }}`
- `src/github/token.ts` — `setupGitHubToken()` returns `OVERRIDE_GITHUB_TOKEN` immediately when set, before ever requesting an OIDC token.

So supplying the input short-circuits the exchange. `on_claude_mention.yml` already relied on exactly this.

## Fix

| Workflow | Token | Why |
|---|---|---|
| `on_issues_ai_triage.yaml` | `GITHUB_TOKEN` | Classify is read-only (`--allowedTools` is limited to `Read,Write,Glob,Grep,Bash(gh issue list:*),Bash(gh search issues:*)`); least privilege, and matches the `GH_TOKEN` the step already exports |
| `on_issue_approved.yaml` | `AI_PR_TOKEN` | Opens PRs — needs a PAT so the PR triggers CI |
| `on_pr_coderabbit.yml` | `AI_PR_TOKEN` | Pushes to PR branches |
| `daily_ai_fix.yaml` | `AI_PR_TOKEN` | Opens PRs |

Both secrets already exist in the `CR_PAT` environment, which all four jobs declare.

Also corrected the stale `id-token: write` comments — the permission is kept as the fallback auth path, but it is no longer the active one.

## Blast radius beyond the reported failure

- `on_pr_coderabbit.yml` was almost certainly broken the same way — `github.actor` there is `coderabbitai[bot]`. Every recent run was `skipped`, so it had not been exercised.
- `daily_ai_fix.yaml` passed only incidentally: it is `schedule`-triggered, so the actor is the maintainer. A `workflow_dispatch` by anyone else would have hit the same 401.

## Verification

All four files pass `yaml.safe_load`. Note that `issues`-event workflows only run from the default branch, so triage stays broken until this merges.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of automated issue triage, approved-plan execution, and pull request follow-ups by aligning authentication behavior and avoiding authorization-related failures.
  * Ensured automated fixes re-trigger the appropriate CI checks after workflow-driven updates.
* **Chores**
  * Updated workflow notes and permissions guidance to clarify when token-based authentication is used versus fallback behavior.

### DIFF
--- a/.github/workflows/daily_ai_fix.yaml
+++ b/.github/workflows/daily_ai_fix.yaml
@@ -50,7 +50,6 @@ permissions:
   contents: write
   issues: write
   pull-requests: write
-  id-token: write  # unused while github_token is set below; kept as the fallback auth path
 
 concurrency:
   group: ai-fix-sweep
--- a/.github/workflows/daily_ai_fix.yaml
+++ b/.github/workflows/daily_ai_fix.yaml
@@ -50,7 +50,7 @@ permissions:
   contents: write
   issues: write
   pull-requests: write
-  id-token: write  # claude-code-action fetches a GitHub OIDC token to auth the OAuth flow
+  id-token: write  # unused while github_token is set below; kept as the fallback auth path
 
 concurrency:
   group: ai-fix-sweep
@@ -129,6 +129,11 @@ jobs:
         uses: anthropics/claude-code-action@44423bdec74b97d67543eb16c110546762c110b2  # v1
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+          # Skip the OIDC -> Claude App token exchange. The scheduled path
+          # happens to pass it (github.actor is the maintainer), but
+          # workflow_dispatch by anyone else would 401. AI_PR_TOKEN, not
+          # GITHUB_TOKEN, so a PR Claude opens triggers CI.
+          github_token: ${{ secrets.AI_PR_TOKEN }}
           # One sticky, auto-updating status comment per run instead of the
           # model narrating its own progress in scattered comments.
           track_progress: true
--- a/.github/workflows/on_issue_approved.yaml
+++ b/.github/workflows/on_issue_approved.yaml
@@ -32,7 +32,6 @@ permissions:
   contents: write
   issues: write
   pull-requests: write
-  id-token: write  # unused while github_token is set below; kept as the fallback auth path
 
 concurrency:
   group: ai-approve-${{ github.event.issue.number || inputs.issue }}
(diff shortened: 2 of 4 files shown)


alexbelgium:
@coderabbitai Addressed in ac02bf2 — removed the `id-token: write` grant outright from all four workflows rather than scoping it to the Claude job.

Verified it is genuinely dead rather than a fallback before removing. Grepped every `.ts` in the action at the pinned SHA `44423bd` for `getIDToken`:

- `src/github/token.ts` — only reached *after* the `OVERRIDE_GITHUB_TOKEN` early return, so unreachable once `github_token` is set
- `base-action/src/workload-identity.ts` — GCP/AWS workload-identity federation; unused here (no Vertex/Bedrock config)
- `base-action/src/parse-sdk-options.ts` — *deletes* `ACTIONS_ID_TOKEN_REQUEST_URL/TOKEN` so the Claude session cannot mint tokens; no dependency on the permission

So the earlier "kept as the fallback auth path" comment was simply wrong, and job-scoping would have preserved a permission nothing can reach.

Removing it also improves the failure mode: if `AI_PR_TOKEN` ever goes missing, `github_token` resolves empty, the action falls through to OIDC, and it now fails with `Could not fetch an OIDC token. Did you remember to add \`id-token: write\`` — instead of the misleading `401 User does not have write access` this PR set out to fix.

Left `on_claude_mention.yml:57` untouched: its grant is already job-scoped, and it is outside this PR's scope.

PR #2922, opened 2026-07-28
fix(claude_desktop): persist sign-in by opting into Electron safeStorage

alexbelgium:
## Problem

Claude Desktop asks the user to sign in again on **every** start.

The v1.35 fix was **inert**. `--password-store=basic` *was* reaching the process (confirmed on a live install's `/proc/<pid>/cmdline`), yet the app still logged, on every launch:

```
[warn] safeStorage not available, tokens will not persist
[warn] Encryption not available, returning empty env vars
[error] Electron safeStorage encryption is not available on this system, cannot store allowlist cache
```

## Root cause

Electron refuses its own `basic_text` backend unless the **application** calls `safeStorage.setUsePlainTextEncryption(true)` before the `ready` event. Claude Desktop never calls it — the symbol is present in the shipped Electron binary but absent from `resources/app.asar`. So `isEncryptionAvailable()` stayed `false` and the auth token was never persisted. v1.35 replaced one unavailable backend with another.

Verified against a standalone Electron of the same generation, all with `--password-store=basic`:

| case | `isEncryptionAvailable()` |
| --- | --- |
| no opt-in (= Claude Desktop as shipped) | `false` |
| `setUsePlainTextEncryption(true)` | `true`, `encryptString` works |
| fresh process decrypting the earlier process's blob | `true`, plaintext recovered |

The third row is the restart survival this add-on needs.

## Alternatives ruled out

- **gnome-keyring / any Secret Service** — its first-boot password prompt blocks Claude Desktop from launching at all, which is a worse failure than a lost session. Out of scope by maintainer decision.
- **A command-line switch** — none exists; the opt-in is API-only.
- **`NODE_OPTIONS=--require` preload** — verified against the real binary: packaged Electron apps ignore `NODE_OPTIONS`, the probe never ran.

That leaves injecting the call into the app's own main bundle.

## The fix

- `rootfs/usr/local/bin/claude-safestorage-patch.js` (new) injects the opt-in into `.vite/build/index.pre.js` inside `app.asar`.
- `rootfs/etc/cont-init.d/86-claude_safestorage.sh` (new) runs it every boot, **after** `81-claude_update.sh` — an apt upgrade of `claude-desktop` ships a fresh unpatched `app.asar`, so a build-time patch would be undone by the first update. Marker-guarded, so an unchanged app is a no-op.

Two details that would otherwise break silently:

- The injection goes **after** the bundle's leading `"use strict";`. A directive prologue only takes effect as the first statement, so prepending would drop the entire main process out of strict mode.
- The archive is **rebuilt**, not edited in place — asar headers store a byte offset and length per file. `unpacked` and symlink entries are carried through untouched (preserving the `app.asar.unpacked` layout) and the per-file SHA-256 `integrity` record is recomputed for the one changed file.

Safety: the patcher **fails closed** — anything not opening with an unambiguously terminated `"use strict"` directive is refused rather than patched by guesswork (`"use strict" + x` is an expression; injecting into it would be a syntax error). The rebuild is written to a temp file, `fsync`ed, and fully re-validated from disk (entry count, every offset/length as an exact in-bounds integer, last body ending precisely at EOF, recomputed hash of the patched entry) before being renamed into place; the temp file is removed on every failure path. A failure is logged and swallowed, under a 120s `timeout`, so it can never block startup — an unpatched app still runs, it just forgets the sign-in.

## Verification

- Byte-for-byte against the real archive: same 211 entries, no `unpacked`/symlink flag drift, no trailing slack, every integrity hash valid, content differing in **exactly one** file.
- Patched bundle passes `node --check`; re-running is a no-op; no temp files left behind.
- Unit tests for the injector cover semicolon / single-quote / ASI-terminated directives (patched, valid syntax) and expression / leading-comment / no-directive forms (correctly refused), plus `integrityOf` for empty and exact-multiple-of-blockSize files.
- `shellcheck` and `node --check` clean.

Reviewed independently by OpenAI Codex (gpt-5.6-sol); its substantive findings — empty-file integrity blocks, ASI-unsafe directive matching, fail-closed injection, short-write loop, insufficient pre-rename verification, `fsync`, temp-file cleanup, offset validation, startup timeout — are all addressed in this branch.

## Note for reviewers

**Not verified:** the patched archive actually running inside Claude Desktop. The mechanism is proven on standalone Electron and the rebuild is proven byte-exact, but the two have not been exercised together, because restarting the app would have killed the session doing the work.

Users need **one** sign-in after upgrading; the previously stored session is already stale.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the recurring “For your security, sign in again…” sign-in prompt on Linux.
  * Improved sign-in persistence for `--password-store=basic` after a one-time re-sign-in post-upgrade.
  * If the persistence adjustment can’t be applied, startup proceeds normally and logs the outcome.
* **Documentation**
  * Updated sign-in documentation to reflect what actually resolved the Linux streamed-desktop OAuth/session issue.
* **Chores**
  * Updated the add-on version to **1.37**.


alexbelgium:
Both review findings addressed in 31702cf:

- **Codex bot (P2), stale temp archive after a timeout** — valid. `timeout` kills the patcher outright, so its own cleanup path never runs, and because the live archive stays unpatched every later boot would retry under a new pid and strand another archive-sized file. The hook now sweeps `.app.asar.addon-tmp.*` itself on the failure path.
- **CodeRabbit, script can exit non-zero** — valid, and it contradicted the file's own comment. The logging `while` loop was the last command, so an empty last line made the script exit 1 and fail cont-init. Added the explicit `exit 0`, and switched the loop body from `[ -n "$line" ] && …` to an `if`, since under `set -e` a false `&&` list is itself enough to abort.

Verified with a harness that runs the real hook against a stub patcher for three cases — fails fast, hangs until `timeout` SIGKILLs it, succeeds. All three now exit 0 and leave no temp files. (My first attempt at that harness was silently empty and 'passed' vacuously; the rebuilt one genuinely exercises the script.)

PR #2956, opened 2026-08-10
fix(stargazer-map): resolve countries via ISO code instead of English name (Russia/Turkey/Ivory Coast were dropped)

alexbelgium:
## Russia renders grey on the stargazer map, and always has

`.github/stargazer_countries.csv` currently holds **963 users with a resolved country**. Russia
appears **zero** times. Turkey appears once — spelled `Türkiye` (row 229), i.e. it resolved only
because Nominatim happened to emit the exact ISO spelling that one time. For scale, the same cache
has Poland 28, Ukraine 12, Kazakhstan 2, Belarus 1. The largest landmass on the map draws grey.

The cause is a vocabulary mismatch. `username_to_country()` geocoded the user's free-text location
and then matched Nominatim's **English display name** against `pycountry`:

```python
for part in reversed(g.raw["display_name"].split(",")):
    country = pycountry.countries.lookup(part.strip()).name   # LookupError -> try next part
```

Those two vocabularies disagree, so some countries can never resolve:

| `pycountry.countries.lookup(...)` | result |
|---|---|
| `"Russia"` | `LookupError` |
| `"Turkey"` | `LookupError` |
| `"Ivory Coast"` | `LookupError` |

pycountry's ISO names are `Russian Federation`, `Türkiye`, `Côte d'Ivoire`; Nominatim with
`language="en"` returns the common names. Affected users were silently recorded as unknown.

## The fix

Ask Nominatim for structured address details and read the ISO 3166-1 alpha-2 code, instead of
matching English prose:

```python
g = GEOL.geocode(loc, language="en", addressdetails=True, timeout=10)
code = ((g.raw.get("address") or {}).get("country_code") or "") if g else ""
country = pycountry.countries.get(alpha_2=code.upper()) if code else None
return country.name if country else ""
```

`addressdetails` is a first-class geopy parameter — **no new dependency**, and still one geocode
request per user. Net **−5 lines**, confined to one function. It also deletes the reversed-component
loop, which was a latent false-positive source: any city or region component that happened to spell
a country name would have matched.

**The return contract is unchanged** — still a pycountry `.name` string. The CSV keeps storing
country names and the renderer keeps resolving them with `pycountry.countries.lookup(...).alpha_3`.
Nothing else in the script is touched.

## Verified

Live Nominatim, single-threaded, 1.2 s between requests, descriptive User-Agent. Old and new logic
read the same geocode response, so the comparison is apples to apples:

| location | Nominatim tail | OLD | NEW |
|---|---|---|---|
| `Moscow` | Russia | *(blank)* | **Russian Federation** |
| `Istanbul` | Turkey | *(blank)* | **Türkiye** |
| `Abidjan` | Ivory Coast | *(blank)* | **Côte d'Ivoire** |
| `Praha` | Czechia | Czechia | Czechia |
| `Wien, Österreich` | Austria | Austria | Austria |
| `Cologne, Germany` | Germany | Germany | Germany |
| `San Francisco, CA` | United States | United States | United States |
| `London` | United Kingdom | United Kingdom | United Kingdom |
| `Seoul` | South Korea | Korea, Republic of | Korea, Republic of |
| `Taipei` | Taiwan | Taiwan, Province of China | Taiwan, Province of China |
| `Tehran` | Iran | Iran, Islamic Republic of | Iran, Islamic Republic of |
| `Hanoi` | Vietnam | Viet Nam | Viet Nam |
| `La Paz, Bolivia` | Bolivia | Bolivia, Plurinational State of | Bolivia, Plurinational State of |
| `Caracas` | Venezuela | Venezuela, Bolivarian Republic of | Venezuela, Bolivarian Republic of |
| `Dar es Salaam` | Tanzania | Tanzania, United Republic of | Tanzania, United Republic of |
| `Chisinau` | Moldova | Moldova, Republic of | Moldova, Republic of |
| `/dev/null` | *(no match)* | *(blank)* | *(blank)* |
| `127.0.0.1` | *(geocoder error)* | *(blank)* | *(blank)* |

Three countries fixed, **no regressions** — every location the old path already resolved resolves
to the identical string.

The above ran the extraction logic standalone. I then re-ran the **actual edited function** with only
the GitHub profile request stubbed and the Nominatim call live, confirming `Moscow -> Russian
Federation`, `Istanbul -> Türkiye`, `Abidjan -> Côte d'Ivoire`, `Cologne, Germany -> Germany`,
`London -> United Kingdom`, and blank for both an empty location and `/dev/null`.

Renderer invariant checked exhaustively: for **all 249** countries in pycountry,
`countries.lookup(c.name).alpha_3 == c.alpha_3` — zero failures. So every name this function can now
emit is one the rendering side can resolve.

Also `python -m py_compile` clean and `black --check` reports no change needed.

## Not verified / assumed

- **How many real stargazers this recovers is unknown.** I did not re-run the full sweep (it would
  rewrite the committed CSV and PNG and take ~1 request/second across the blank rows). That the
  three countries now resolve is *verified*; the recovered-user count is *assumed non-zero*.
- The committed `.github/stargazer_countries.csv` and `.github/stargazer_map.png` are deliberately
  **untouched** in this PR. The next scheduled workflow run regenerates them.
- One honest caveat, **pre-existing and unchanged by this PR**: junk locations that happen to name a
  real US place still resolve. `Earth` (Earth, Texas) and `Remote` (Remote, Oregon) both returned
  `United States` under the old code *and* the new code. This PR neither introduces nor fixes that;
  filtering it would need new heuristics, which is out of scope here.

## Overlap with in-flight work

Two other PRs touch this same file. **This PR does not conflict semantically with either** — it
changes only the body of `username_to_country()`, but expect a textual rebase.

1. **#2953** (`feat/stargazer-map-stats`) modifies the rendering half. No overlap with this function.
2. **The cache PR** adds a `last_checked` column; rows with an empty `last_checked` count as
   never-checked and are re-queried on its first run. Since the Russian/Turkish/Ivorian users are
   currently stored blank, they are re-queried — and with this fix in place, resolved — on that run.

Merge order does not matter.

## Rollback

Single-hunk, single-function change — revert the commit and the previous behaviour returns exactly.
No schema, cache-format, dependency, or workflow change to unwind. `.github/workflows/generate_stargazer_map.yml`
is untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved location country detection using standardized country codes.
  * Increased reliability when location details are incomplete or formatted differently.
  * Ignored known non-location entries during location lookup.
  * Preserved valid locations, including entries such as “Earth, TX.”
  * Geocoding failures and unrecognized country codes continue to return no country.

### DIFF
--- a/.github/generate_map.py
+++ b/.github/generate_map.py
@@ -106,20 +106,15 @@ def username_to_country(login):
     if not loc.strip():
         return ""
     try:
-        g = GEOL.geocode(loc, language="en", timeout=10)
+        g = GEOL.geocode(loc, language="en", addressdetails=True, timeout=10)
     except Exception:
         return ""
-    if not g or "display_name" not in g.raw:
-        return ""
-    # take the last comma-separated component that matches a country
-    for part in reversed(g.raw["display_name"].split(",")):
-        part = part.strip()
-        try:
-            country = pycountry.countries.lookup(part).name
-            return country
-        except LookupError:
-            pass
-    return ""
+    # Use the ISO code from the structured address: Nominatim's English display
+    # names ("Russia", "Turkey", "Ivory Coast") do not all match pycountry's ISO
+    # names ("Russian Federation", "Türkiye", "Côte d'Ivoire").
+    code = ((g.raw.get("address") or {}).get("country_code") or "") if g else ""
+    country = pycountry.countries.get(alpha_2=code.upper()) if code else None
+    return country.name if country else ""
 
 
 def count_by_country(cache):
--- a/.github/generate_map.py
+++ b/.github/generate_map.py
@@ -61,6 +61,35 @@ HEADERS = {
 }
 GEOL = Nominatim(user_agent="gh-stargazer-map")
 
+# Non-answers that Nominatim happily resolves to a real place: "Earth" is a
+# town in Texas, "Remote" is a settlement in Oregon.  Matched on the whole
+# stripped, lowercased string only -- "Earth, TX" is someone's actual address
+# and must still geocode.
+JUNK_LOCATIONS = {
+    "127.0.0.1",
+    "/dev/null",
+    "anywhere",
+    "earth",
+    "everywhere",
+    "here",
+    "home",
+    "internet",
+    "localhost",
+    "mars",
+    "moon",
+    "n/a",
+    "none",
+    "nowhere",
+    "null",
+    "planet earth",
+    "remote",
+    "space",
+    "the internet",
+    "unknown",
+    "world",
+    "worldwide",
+}
+
 
 # -----------------------------------------------------------------------------
 
@@ -104,6 +133,8 @@ def username_to_country(login):
     resp.raise_for_status()
     loc = (resp.json() or {}).get("location") or ""
     if not loc.strip():
+        return ""
+    if loc.strip().strip(".!").lower() in JUNK_LOCATIONS:
         return ""
     try:
         g = GEOL.geocode(loc, language="en", addressdetails=True, timeout=10)


alexbelgium:
Pushed `c51423e` — the junk-location filter the review flagged as a known gap is now closed.

```python
JUNK_LOCATIONS = {"127.0.0.1", "/dev/null", "anywhere", "earth", "everywhere", "here",
                  "home", "internet", "localhost", "mars", "moon", "n/a", "none",
                  "nowhere", "null", "planet earth", "remote", "space", "the internet",
                  "unknown", "world", "worldwide"}
```

Checked **before** the geocode call, so these also stop costing a Nominatim request.

The important design point: the match is on the **whole** stripped/lowercased string, never a substring. `Earth` is a town in Texas and `Remote` is a settlement in Oregon, so a substring match would wrongly discard people who actually live there.

Verified with the network stubbed (fake geocoder that records whether it was called):

| input | geocoder called | result |
|---|---|---|
| `Earth`, `earth`, `REMOTE`, `  Remote  `, `Earth.` | no | `""` |
| `Worldwide`, `127.0.0.1`, `/dev/null`, `The Internet`, `n/a` | no | `""` |
| `Earth, TX`, `Remote, Oregon` | **yes** | geocoded normally |
| `Moscow`, `Berlin, Germany`, `Home Counties, UK`, `Northern Mars, PA` | **yes** | geocoded normally |

All 16 cases behaved as expected. `black --check` clean.

Not verified: how many of the 1689 currently-blank users this actually affects — they are blank either way, so this changes wasted Nominatim requests and future correctness, not the current map.

PR #2965, opened 2026-08-12
Obsidian Sync Server add-ons (CouchDB LiveSync backend) - three flavours

ToledoEM:
Three add-ons that run CouchDB as a backend for the [Obsidian Self-hosted LiveSync](https://github.com/vrtmrz/obsidian-livesync) plugin. Vaults sync between a user's own devices through Home Assistant, with no Obsidian Sync subscription and no notes leaving the local network.

Plain CouchDB is not usable as a LiveSync backend out of the box. The plugin needs a single-node cluster, CORS opened to Obsidian's app origins, authentication required, and raised request and document size limits. Each add-on applies that configuration on every start, idempotently, so it also repairs settings someone changed by hand in Fauxton. The values come from the upstream provisioning tool (`vrtmrz/obsidian-livesync`, `utils/couchdb/provision.ts`).

All three are the same server. **They differ only in how TLS gets in front of it.**

## Add-ons

| slug | TLS | Ports | Use when |
|---|---|---|---|
| `obsidian_syncserver_solo` | none | 5984 | A reverse proxy already exists |
| `obsidian_syncserver_ssl` | CouchDB serves HTTPS from certs in `/ssl` | 5984, 6984 | Certificates are on the HA machine |
| `obsidian_syncserver_npm` | bundled Nginx Proxy Manager | 5984, 80, 81, 443 | No proxy, want cert handling included |

Desktop Obsidian is happy over plain HTTP. Mobile Obsidian is not, since it insists on a valid TLS certificate. That is the only reason the `ssl` and `npm` flavours exist.

## What they all do

Each one creates the vault database if it is missing, treating a 412 on restart as "already there" rather than an error. If the password option is left blank it generates a strong one and writes it to `/config/obsidian-syncserver/admin_password`, then reuses it on later starts. Regenerating it every restart would invalidate credentials the user already typed into Obsidian.

Data lives under `/config/obsidian-syncserver/data`, not `/data`. `/data` gets wiped on reinstall and is not part of a Home Assistant backup, and the vault is the entire point of the add-on.

The database name is validated up front. Skip that and a bad name fails much later at the create step, with an opaque 400 that tells the user nothing.

`obsidian_syncserver_ssl` also checks the certificate before starting: file present, valid PEM, not expired, key actually matches. It refuses to start otherwise and says which of those failed. A broken certificate would otherwise reach the user as a connection failure with no explanation, which is miserable to debug. On a good start it logs the hostnames covered and the expiry date.

`obsidian_syncserver_npm` symlinks `/etc/letsencrypt` to `/data` from an s6-rc oneshot. NPM's own `prepare` service aborts the whole boot when `/etc/letsencrypt` is not a mount, and Home Assistant does not mount it.

## Notes for review

Base images: `solo` and `ssl` build from `couchdb:3.5.2`. `npm` is a multi-stage build that lifts `/opt/couchdb` out of `couchdb:3.5.2` into `jc21/nginx-proxy-manager`. That works because the CouchDB release is self-contained, including its own Erlang runtime, and both images are Debian trixie. `libicu76` and `libmozjs-128-0` are installed explicitly for it. I pinned the NPM base to `2.15.1` instead of `:latest`, since a floating base would let that ABI assumption break silently on some future rebuild.

No `.templates/` usage. All three build from a pinned upstream image without `ha_automodules.sh` or `ha_entrypoint.sh`, the same as `filebrowser`, `autobrr`, `photoprism` and `joplin`.

s6 layout differs between them. `solo` and `ssl` run `run.sh` directly as the entrypoint, because the official CouchDB image is Debian-based and carries no s6-overlay at all. `npm` inherits s6-overlay v3 from the NPM image and uses `s6-rc.d` services.

**Updater**: `solo` and `ssl` track `library/couchdb` on Docker Hub. `npm` is `paused: true` because it tracks two upstreams that the updater cannot resolve together, so it gets bumped by hand.

Versions: `solo` and `ssl` at `3.5.2`, `npm` at `3.5.2.1` for the base image pin. Dot, not hyphen, so Supervisor does not read it as a pre-release and quietly skip the update.

## Submitter list

This also adds me to `COMMUNITY_SUBMITTED_ADDONS.md` and `.github/addon_submitters.json` for the three Obsidian add-ons, plus `bentopdf` and `nginx_webserver_proxy`, which were already in the repo but had no table entry.

Both files now use add-on slugs as the key, so they line up. The old `nginx proxy manager` key is renamed to `nginx_webserver_proxy` to match. Note that `on_issues_ping_submitter.yml` matches keys as whitespace-delimited substrings of the issue title and body, so with a slug key the ping fires when someone writes the slug, not when they write the display name. Say the word if you would rather keep display-name keys for that reason.

Each README also gets a Support section pointing issues at me, worded the same as my existing add-ons, with links to the LiveSync plugin and CouchDB upstreams. The `npm` one spells out that it covers the add-on rather than CouchDB or Nginx Proxy Manager themselves, since it bundles both and misrouted issues are likely.

## Testing

- Linted locally, all three clean: `bash -n`, ShellCheck, Hadolint. `config.yaml` parses, every `options` key has a matching `schema` entry and a `translations/en.yaml` entry, and `updater.json` is valid JSON. `.github/addon_submitters.json` still parses after the edit.

- All three were built and installed on a live HAOS machine (aarch64) and sync a real vault with the Self-hosted LiveSync plugin. That covers the `npm` multi-stage build, which was the part I was least sure of, since it assembles an image out of two unrelated bases.

- Each one reaches `Ready. Point Obsidian Self-hosted LiveSync at this server.` in the log, writes `/config/obsidian-syncserver/admin_password`, and answers `curl -u admin:<pw> http://<host>:5984/obsidian` with the database document. Restarts are clean, so the idempotent provisioning holds.

- amd64 has not been tested on hardware, so CI is the gate for that arch.

## Summary by CodeRabbit

* **New Features**
  * Added three Obsidian Sync Server add-ons with standalone, SSL-enabled, and Nginx Proxy Manager configurations.
  * Added persistent CouchDB storage, automatic database setup, credential handling, optional TLS, and LiveSync support.
  * Added BentoPDF and Nginx-related community add-on entries.

* **Documentation**
  * Added installation, configuration, security, troubleshooting, backup, and connection guidance.
  * Added changelogs and release metadata for the new add-ons.


ToledoEM:
Because of https://github.com/alexbelgium/hassio-addons/issues/2966  I did this too https://github.com/alexbelgium/hassio-addons/pull/2965/commits/f5eccdec4b6677b005b1b401d8e375b9c8fc94e4 

You can pick which one to merge

alexbelgium:
Hi, sorry got a bit overwhelmed by the activity, should I merge this PR ?

ToledoEM:
Ai noise, totally an issue. I will revert the changes and will be ready to merge :-) 

PS: There are ways to setup obsidian in a docker, but they involve a full linux install desktops with the app installed, since there are no web versions for obsidian. I still like it

Issue #1822, opened 2025-03-27
❓ [BirdNET-Pi Docker Standalone] Can't find the USB audio device in Respberry Pi 4B


usersync:
Thank you for making the BirdNET-Pi standalone docker.

**Problem Description** 

When I try to run it on my Respberry Pi 4B, the docker start log showed me that it can not find the USB audio card.
```
+ sudo systemctl restart birdnet_recording.service
ALSA lib confmisc.c:855:(parse_card) cannot find card '0'
ALSA lib conf.c:5180:(_snd_config_evaluate) function snd_func_card_id returned error: No such file or directory
ALSA lib confmisc.c:422:(snd_func_concat) error evaluating strings
ALSA lib conf.c:5180:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1334:(snd_func_refer) error evaluating name
ALSA lib conf.c:5180:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:5703:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2666:(snd_pcm_open_noupdate) Unknown PCM dsnoop
arecord: main:831: audio open error: No such file or directory
+ for i in "${services[@]}"
+ sudo systemctl restart birdnet_analysis.service
+ sudo systemctl restart birdnet_log.service
[server][INFO] LOADING TF LITE MODEL...
INFO: Created TensorFlow Lite XNNPACK delegate for CPU.
[server][INFO] LOADING DONE!
2025/03/27 16:10:10 GoTTY is starting with command: cat /proc/1/fd/1
2025/03/27 16:10:10 HTTP server is listening at: http://127.0.0.1:8080/log/
[birdnet_analysis][INFO] backlog is 0
[birdnet_analysis][INFO] backlog done
+ sudo systemctl restart birdnet_stats.service
+ for i in {1..5}
+ systemctl is-active --quiet birdnet_analysis.service
+ logger '[/home/pi/BirdNET-Pi/scripts/restart_services.sh] birdnet_analysis.service is running'
+ break
... starting livestream services
  You can now view your Streamlit app in your browser.
  URL: http://localhost:8501/stats
Failed to create secure directory (/run/user/1000/pulse): No such file or directory
ALSA lib confmisc.c:855:(parse_card) cannot find card '0'
[...]

PR #2974, opened 2026-08-15
fix(birdnet-pi-zach): make ALSA_CARD actually select the microphone

alexbelgium:
Ports the `ALSA_CARD` fix from #2972 to the sibling add-on, which carries the identical defect.

## Why this is an exact port, not an adaptation

`birdnet-pi-zach` wraps the [zach7036 BirdNET-Pi fork](https://github.com/zach7036/BirdNET-Pi-Enhanced-Version), but the add-on layer is shared:

- `99-run.sh` was **byte-identical** to `birdnet-pi`'s before #2972 (`diff` against `origin/master:birdnet-pi/...` returns nothing)
- `01-structure.sh` is **still byte-identical**, so the symlink premise the fix depends on holds unchanged
- the result of this patch is **byte-identical** to the fixed file on `ai-fix/birdnet-pi-1822`

So there is no behavioural guesswork: the same code had the same bug, and now has the same fix.

## The bug

`ALSA_CARD` was copied straight into `REC_CARD`, but BirdNET-Pi hands `REC_CARD` to `arecord -D` and `ffmpeg -f alsa -i`, which expect an ALSA **PCM name**. A card index like `1` or an id like `Audio` is not one — it produces `Unknown PCM 1` and nothing records at all. It is now converted to `plughw:CARD=<value>,DEV=0`, while a value that already is a PCM name (`dsnoop:CARD=Audio,DEV=0`, `default`, `null`, `pulse`, `pipewire`) is passed through untouched.

Second defect in the same block: `sed -i` **replaces a symlink with a regular file**. `01-structure.sh` links `$HOME/BirdNET-Pi/birdnet.conf` -> `/config/birdnet.conf` (same `CONFIG_FILES` loop, confirmed present here), so the old code detached the config from the file the WebUI writes to. It now edits `/config/birdnet.conf` directly with `--follow-symlinks`.

## Version

`2026.07.10.2` -> `2026.08.15`, matching the new CHANGELOG heading.

Not a `.N` counter increment: this add-on's `updater.json` has an empty `upstream_version` and is `paused`, so there is no upstream prefix to measure a counter against. That is the "leave the counter alone, a human decides" case, and the decision here follows the family's date convention — the same one `birdnet-pi` uses, where every version lines up with a dated CHANGELOG heading.

The bump is what makes the fix real: Supervisor only offers a rebuild when `version` changes, so without it this would merge and the add-on would keep running the old image with `ALSA_CARD` still broken.

## Verification

**Verified** — the patched file is byte-identical to the fixed `birdnet-pi` version; `bash -n` parses; `2026.08.15 > 2026.07.10.2` under component-wise comparison, so Supervisor sees it as newer; `config.yaml` parses and `slug` is unchanged.

**Checked** — `shellcheck` reports only two **pre-existing** `SC2015` warnings in the unrelated `AUDIO_GID` block; both are present on `master` and deliberately untouched. The diff is a single hunk confined to the `ALSA_CARD` block.

**Not verified** — the runtime behaviour on real hardware. I have no BirdNET-Pi with a USB microphone to exercise `arecord`, and CI only builds the image. The reasoning is inherited from #2972, where the same change is under review against the original report (#1822).

## Separate PR on purpose

Kept off #2972 rather than appended to it: the repo's own fix rules are one add-on per branch, one branch per pull request, and #2972 is already in review — widening its scope mid-review would surprise its reviewers and mean a failure in either add-on blocks both. Happy to squash them together if you would rather have one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ALSA microphone selection by supporting card indices, IDs, and existing PCM device names.
  * Ensured configuration updates correctly follow symlinks and modify the active configuration file.

* **Chores**
  * Updated the add-on version to `2026.08.15`.
  * Documented the fixes in the changelog.

### DIFF
--- a/birdnet-pi-zach/CHANGELOG.md
+++ b/birdnet-pi-zach/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 2026.08.15 (15-08-2026)
+
 - Fix: `ALSA_CARD` now really selects the microphone. Its value was copied as-is into `REC_CARD`, but BirdNET-Pi hands `REC_CARD` to `arecord -D` / `ffmpeg -f alsa -i`, which expect an ALSA PCM name: a card index such as `1` gave `Unknown PCM 1` and no recording at all. It is now converted to `plughw:CARD=<value>,DEV=0`, while a value that already is a PCM name (`dsnoop:CARD=Audio,DEV=0`, `default`, `null`, `pulse`, `pipewire`, ...) is used as provided
 - Fix: writing `REC_CARD` no longer detaches `birdnet.conf` from `/config`. `sed -i` replaced the `$HOME/BirdNET-Pi/birdnet.conf` symlink with a regular file, so later edits from the WebUI went to a different file than the one the add-on had written; it now edits `/config/birdnet.conf` with `--follow-symlinks`
+
 ## 2026.07.10-2 (10-07-2026)
 - Minor bugs fixed
 ## 2026.07.10 (10-07-2026)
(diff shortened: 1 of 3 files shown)


alexbelgium:
Superseded by #2972, which now bundles the identical birdnet-pi-zach fix (same 99-run.sh change, CHANGELOG entry, and version bump) alongside the birdnet-pi fix it was copied from, plus the matching README_standalone.md doc addition. Closing to avoid two PRs shipping the same change.

PR #2979, opened 2026-08-16
Fix FileBrowser Quantum direct web access

### LINKED ISSUE #2978: 🐛 [FileBrowser Quantum]  No Network section despite README documenting direct access on port 8071
### Description

The FileBrowser Quantum README says that the web interface can be accessed at:

`http://<home-assistant-ip>:8071`

However, the add-on does not display a **Network** section in Home Assistant, and neither of these addresses is reachable:

- `http://homeassistant.local:8071`
- `http://homeassistant.local:8080`

Access through Home Assistant Ingress works correctly.

### Reproduction steps

```bash
1. Go to `http://homeassistant.local:8071`
2. See error
```

### Addon Logs

```bash
2026/08/16 20:13:42 [INFO ] Running at               : http://0.0.0.0:8080/api/hassio_ingress/Xs1tYmjYKlLbQJO4QWDO03LWjxF_gMVVqBPcUkNK-Js/
```

### Architecture

aarch64

### OS

HAos

polmonta:
## Summary
- Publish FileBrowser Quantum's internal `8080/tcp` on host port `8071`.
- Add the port description so Home Assistant renders the Network section.
- Bump the add-on patch version and update the changelog.

## Verification
- Ruby YAML/metadata regression check passed.
- `git diff --check` passed.
- Docker-based add-on linter/build could not be run locally because the Docker daemon is unavailable.

Fixes #2978

## Summary by CodeRabbit

* **New Features**
  * Exposed the FileBrowser web interface on host port 8071.
  * Added direct access with automatic redirection to `/filebrowser_quantum/`.
  * Retained access through Home Assistant Ingress.

* **Documentation**
  * Updated installation and configuration guidance with the new web address.
  * Added release notes for version 1.5.1.1.

### DIFF
--- a/filebrowser_quantum/CHANGELOG.md
+++ b/filebrowser_quantum/CHANGELOG.md
@@ -1,6 +1,9 @@
  
 ## 1.5.1.1 (2026-08-16)
-- Expose the web UI on host port 8071.
+- Expose the web UI on host port 8071, reachable at `<your-ip>:8071`
+  (redirects to `/filebrowser_quantum/`). Direct access is served by a new,
+  separate nginx vhost that proxies to the same backend Ingress already uses;
+  Ingress itself, and the app's own base URL, are unchanged.
 
 ## 1.5.1 (2026-08-08)
 - Update to latest version from gtsteffaniak/filebrowser (changelog : https://github.com/gtsteffaniak/filebrowser/releases)
--- a/filebrowser_quantum/CHANGELOG.md
+++ b/filebrowser_quantum/CHANGELOG.md
@@ -1,4 +1,7 @@
  
+## 1.5.1.1 (2026-08-16)
+- Expose the web UI on host port 8071.
+
 ## 1.5.1 (2026-08-08)
 - Update to latest version from gtsteffaniak/filebrowser (changelog : https://github.com/gtsteffaniak/filebrowser/releases)
  
--- a/filebrowser_quantum/README.md
+++ b/filebrowser_quantum/README.md
@@ -42,11 +42,11 @@ comparison to installing any other Home Assistant add-on.
 1. Click the `Save` button to store your configuration.
 1. Start the add-on.
 1. Check the logs of the add-on to see if everything went well.
-1. Access the web UI through the sidebar or at `<your-ip>:8071`.
+1. Access the web UI through the sidebar or at `<your-ip>:8071/filebrowser_quantum/`.
 
 ## Configuration
 
-The web UI can be found at `<your-ip>:8071` or through the Home Assistant sidebar when using Ingress.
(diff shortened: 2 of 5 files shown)


alexbelgium:
Ah I missed this pr and merged the ai tool one updating documentation is it something needed for you ? I can revert the other

polmonta:
@alexbelgium would be very nice to have. The Filebrowser one already has it and it's very useful if you want to access Filebrowser Quantum directly. In my case, I have a custom url that redirects directly to Filebrowser. With the current implementation, it's not possible to do the same with Filebrowser Quantum. Thanks!

alexbelgium:
Pushed a commit on top of this PR to make direct access actually work, per the analysis in #2978's comment thread: publishing port 8080 alone gives an unreachable page, because the app's `server.baseURL` is set at boot to the Supervisor ingress-entry path (an opaque, per-install hash).

## What changed

- `ports:` now publishes a **new nginx vhost** (internal port 8072) to host 8071, not the app's own port 8080 directly.
- That vhost (`rootfs/etc/nginx/servers/direct.conf`) rewrites the fixed public path `/filebrowser_quantum/` onto the same ingress-entry baseURL the existing ingress vhost already targets — so the app's own `server.baseURL` and the existing ingress vhost are **completely untouched**. Ingress access should be unaffected.
- `<your-ip>:8071` now redirects to `/filebrowser_quantum/`.

## What's verified vs. not

- **Verified:** `bash -n`, shellcheck, hadolint, config.yaml schema, and the nginx `proxy_pass`-with-URI rewrite math (request-side: `/filebrowser_quantum/foo` → `<ingress-hash>/foo` at the backend) — checked twice independently against nginx's documented substitution semantics.
- **Not verified — needs a real instance:** whether FileBrowser Quantum's *responses* (HTML, JS bundle, API/asset/WebSocket URLs) are relative or embed the absolute ingress-hash path. If they're absolute, the browser will request `<ip>:8071/api/hassio_ingress/<hash>/...` directly after the first page load, which the new vhost has no route for and would 404. I don't have a running Supervisor or local dockerd in this environment to check.

## Before merging, please test on a real instance

1. Install/update the add-on, open `http://<ip>:8071/`.
2. Confirm it redirects to `/filebrowser_quantum/` and the page loads.
3. **Open the browser's network tab** and confirm no requests go to `/api/hassio_ingress/...` (those would 404 — if you see them, the fix needs response-body rewriting too, not just request-side).
4. Confirm login, file listing, upload, and (if used) realtime/WebSocket indexing all work.
5. Separately confirm Ingress access via the HA sidebar still works unchanged.

If step 3 shows ingress-hash requests leaking through, the safe fallback is to drop `direct.conf` and instead publish port 8080 directly (this PR's original approach), documenting the *actual* ingress-entry URL from the add-on's own startup log rather than a fixed path — uglier, but correct by construction since there's no rewriting involved.

PR #3014, opened 2026-08-24
docs(skill): simplify again after the code review, and demand a trigger for defensive branches

alexbelgium:
Process fix, prompted by #3013. No add-on is touched.

## What went wrong

#3013 went into review at 25 lines of code and merged at 10. The excess had two sources, and the workflow caught neither.

**The loop had no simplify pass after the review.** The full loop reads `implement → simplify → Codex reviews the code → PR`. Adversarial review is asked to find what could go wrong, so its output is a list of arguments for more code, and it is never asked whether the branch it wants is reachable. Accepting objections therefore only ratchets the diff upward, and nothing downstream walks it back. Two of the three objections on #3013 were correct and cost two tokens on an existing line; the third had already been answered by my own measurement. Nothing in the loop was positioned to notice that.

**The standing rule was written about performance.** It said complexity is bought only by a measurement showing a concrete user-visible cost, "never by reasoning about hypothetical performance". The branch that actually bloated #3013 was written for robustness, not speed — a pure-bash environment writer for images that ship `with-contenv` but not `s6-dumpenv`, reasoned from the two binaries living in different s6 packages. It was never demonstrated on any real image, and the case it defended would have degraded to the pre-fix behaviour anyway. Because it wasn't a performance argument, the rule didn't visibly apply to it.

That fallback then dragged in a helper function and a second reset that existed only to serve it. Deleting it deleted all three.

It took the maintainer asking "is this the simplest way possible?" to run the pass that should have been in the loop.

## What changed

- **Standing rule** now covers hypothetical *hosts* as well as hypothetical performance: a defensive branch is complexity like any other — name the input that reaches it and the image it happens on, or delete it and let the case fail visibly.
- **Step 3** (attack your own plan) asks the same question before any code exists, which is where it is cheapest to answer. This is the half aimed at the *initial* proposal rather than the cleanup.
- **Step 6** now ends by re-running step 5's checks over the hunks the review touched, and says plainly that a fix collapsing back to fewer lines than it had before the review is the normal outcome.
- **`references/codex-review.md`** gains the distinction to draw while reading objections: "this is wrong" is a bug you fix; "this is undefended" is a claim about some host, and needs the same demonstration you would demand of a measurement.
- **`references/simplify.md`** gains the check and #3013 as its fourth case study, alongside the three that were already there.

The full-loop step list is updated to name the extra pass, so it is visible from the summary rather than only inside step 6.

## Verified

`markdownlint-cli2` over the skill tree: 3 findings, all three present identically on `master` before this change. No new findings.

Not verified: whether the added pass actually changes behaviour on the next full-loop task — that is only observable by running one.

### DIFF
--- a/.claude/skills/hassio-addon-workflow/references/simplify.md
+++ b/.claude/skills/hassio-addon-workflow/references/simplify.md
@@ -13,6 +13,15 @@ Being able to build the complicated thing is not a reason to.
 - A resolution cap shipped as a **new init script writing an s6 envdir** — the wrong mechanism
   entirely (ladder level 4). Renaming the option to the env var the service already reads
   (level 1) would have worked, and the new script did not.
+- A `.templates/ha_entrypoint.sh` fix went to review at 25 lines of code and merged at 10. Two
+  sources of the excess, and neither was caught by the loop: a **pure-bash fallback** written at
+  implement time for images shipping `with-contenv` but not `s6-dumpenv` — reasoned from the two
+  binaries living in different s6 packages, never demonstrated on any real image, and the case it
+  defended would have degraded to the pre-fix behaviour anyway — and a **helper function plus a
+  second reset** that existed only to serve that fallback. Deleting the fallback deleted all of
+  it. The rest of the review's objections were correct and cost two tokens on an existing line.
+  It took the maintainer asking "is this the simplest way possible?" to run the pass that step 6
+  now requires.
 
 ## Checks worth running against your own diff
 
@@ -25,6 +34,11 @@ Being able to build the complicated thing is not a reason to.
   A removal that depends on a host default still needs the same verification as an addition.
 - **Is the fix bigger than the thing it fixes?** That is a smell, not a rule — but it usually
   means the problem was framed one level too deep.
+- **For each defensive branch: what input reaches it, on which image or host?** Go and check,
+  the way you would check a measurement. If you cannot produce the case, delete the branch — the
+  situation then fails the way it already fails today, visibly, instead of through a second path
+  that is never exercised and silently rots as the base images move. Write down in the PR body
+  what you cut and why, so the next person does not re-add it from the same reasoning.
 - **How does this fail in three years**, when the base image, Electron, or upstream has moved?
   Code that reads a documented knob keeps working. Code that reaches into private internals
   does not.
--- a/.claude/skills/hassio-addon-workflow/SKILL.md
+++ b/.claude/skills/hassio-addon-workflow/SKILL.md
@@ -29,8 +29,8 @@ Triage first, then one of two paths:
   resolve bot comments.
 - **Full loop** — performance/RAM/CPU work, diagnosis, anything changing a shipped default,
   ladder levels 4-6, or an explicit Codex-check request: scope → measure → plan → Codex reviews
-  the plan → implement → simplify → Codex reviews the code → PR → resolve comments → verify in
-  production → report.
+  the plan → implement → simplify → Codex reviews the code → **simplify again** → PR → resolve
+  comments → verify in production → report.
 
 Escalate mid-flight if a light task grows — touches a default, needs a new script or service, or
 reveals a deeper problem.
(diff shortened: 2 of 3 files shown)


.claude/skills/hassio-addon-workflow/references/simplify.md
```diff
@@ -25,6 +34,11 @@ Being able to build the complicated thing is not a reason to.
   A removal that depends on a host default still needs the same verification as an addition.
 - **Is the fix bigger than the thing it fixes?** That is a smell, not a rule — but it usually
   means the problem was framed one level too deep.
+- **For each defensive branch: what input reaches it, on which image or host?** Go and check,
+  the way you would check a measurement. If you cannot produce the case, delete the branch — the
```
alexbelgium, line 38: **This contradicts the `/dev/shm` bullet two entries above it, and the stronger rule loses.**

Bullet 2 of this same list says a removal that depends on a host default needs the same verification as an addition, citing `/dev/shm` as a removal that reintroduced a crash loop. `traps.md` is more explicit still: "keep the flag when the size cannot be determined, because the crash it prevents is worse than its overhead."

`--disable-dev-shm-usage` is exactly a defensive branch whose triggering case cannot be produced on the host you are standing on — Home Assistant ignores the add-on's `shm_size`, so the size varies per install and measured 7.7 GB here. Read literally, "if you cannot produce the case, delete the branch" deletes it, and the skill then contains the instruction that caused one of its own documented regressions.

What actually separates the two cases is not reproducibility but whether the case can be **named**: nobody could name a single real image shipping `with-contenv` without `s6-dumpenv` (#3013's fallback), whereas Docker's 64 MB `/dev/shm` default is documented upstream behaviour that HA does not override — named, just not instantiated here. Cost asymmetry decides the rest: a one-flag guard against a crash you cannot rule out is cheap, a second code path that degrades to the pre-fix behaviour anyway is not.

Suggest requiring a *named* real image or host rather than a produced one, and cross-referencing the `/dev/shm` bullet so the two are read together.

---
_Generated by [Claude Code](https://claude.ai/code)_

alexbelgium (commented):
Reviewed under the workflow skill this PR edits. The diagnosis is right and the second simplify pass is the correct place to put it — adversarial review does only argue for more code, and nothing downstream walked it back. Four findings, one of them blocking in my view.

**1. The new defensive-branch rule contradicts the `/dev/shm` bullet two entries above it** (inline, `simplify.md:38`). Read literally it deletes `--disable-dev-shm-usage`, which `traps.md` says to keep precisely *because* the triggering case cannot be determined locally. The distinction that reconciles them is **named** versus **produced**, plus cost asymmetry. Detail inline. This is the one I would fix before merging, because the skill would otherwise carry an instruction that caused one of its own documented regressions.

**2. Codex's open P2 is correct and unaddressed.** The new pass edits code after step 4's `validate.sh` and behavioural tests have already run, and nothing re-runs them; a fix that "collapses back to fewer lines" is exactly the kind of edit that can reintroduce a syntax error. One clause in step 6 closes it — re-run `validate.sh <addon> --vs-master` and the behavioural tests after the second pass, not only after step 4.

**3. `simplify.md` line 4 still reads "In all three cases" while the list now has four case studies.** Not in this diff, so I could not anchor it inline. #3015 rewrites that same sentence to "In each case"; whichever PR merges second should carry the fix.

**4. This PR and #3015 conflict with each other, though both are clean against master.** Verified by merging the two branches locally: `CONFLICT (content): .claude/skills/hassio-addon-workflow/references/simplify.md`. Both append a case study immediately after the resolution-cap bullet (#3014 adds `ha_entrypoint.sh`/#3013, #3015 adds calibre-web/#3010). Resolution is trivial — keep both bullets, they are independent lessons — but the second merge will need a hand.

One observation rather than a finding: the same lesson now appears in the standing rule, step 3, step 6, `codex-review.md` and `simplify.md`. Each sits at a genuinely different point of use, so I would not cut any of them, but it is worth noticing that a PR about complexity ratcheting added +38 lines of process. The step 6 paragraph is the one that changes behaviour; the rest is reinforcement.

Not verified: whether the added pass changes behaviour on a real full-loop task — as your PR body says, only running one shows that. The `codex` CLI is not installed in this container, so I could not run an independent `gpt-5.6-sol` pass locally; the Codex connector's own review of `d9b92f4` is what finding 2 refers to, and a fresh `@codex review` after any fix would re-run it against the new head.

---
_Generated by [Claude Code](https://claude.ai/code)_

alexbelgium:
Closing in favour of #3015, which now carries this PR's changes in full.

The two PRs edited the same three skill files and conflicted in `references/simplify.md` — both appended a case study at the same spot — so consolidating into one avoids a hand-resolved merge for whichever landed second. Everything here is carried over unchanged: the second simplify pass in the full loop, the standing rule covering hypothetical hosts, step 6's re-run of step 5's checks, the `codex-review.md` wrong-vs-undefended distinction, and the `ha_entrypoint.sh` case study. Both case studies are kept side by side.

Three things were applied on top, from the review on this PR:

- The defensive-branch rule contradicted the `/dev/shm` bullet two entries above it in the same list. The bar is now naming the case rather than producing it, so the `/dev/shm` guard stays (documented Docker default HA does not override) while the `s6-dumpenv` fallback still goes (no image could be named).
- Codex's P2 here is closed: step 6 re-runs `validate.sh` and the behavioural tests over the final diff, since the second pass edits code after step 4's checks already ran.
- Step 3's bullet lost two lines that restated the standing rule verbatim, and `simplify.md` no longer says "In all three cases" while listing five.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3051, opened 2026-09-07
add option to configure nfs protocols so user can decide which to be checked

morgaroth:
## Summary by CodeRabbit

* **New Features**
  * Added an optional `nfsversions` setting for configuring the NFS protocol versions attempted when mounting shares.
  * Supports comma-, space-, or newline-separated version lists.
  * Existing installations retain the default fallback order of 4.2, 4.1, 4, and 3 when unset.

* **Documentation**
  * Updated release changelogs across supported add-ons to document the new setting.

* **Chores**
  * Updated add-on versions to reflect the release.


alexbelgium:
Hi, thanks for the proposal. Which issue is it solving ? Isn’t the automatic test enough to find the right protocol ? The more options we add the more we increase complexity for the end user and risk having things messed up and issues coming up to justify ;) thanks

morgaroth:
> Hi, thanks for the proposal. Which issue is it solving ? Isn’t the automatic test enough to find the right protocol ? The more options we add the more we increase complexity for the end user and risk having things messed up and issues coming up to justify ;) thanks

well it is seems it is not enough to observe successful mount

the mount completes on 4.2, but then when trying to read a song (i observed it on navidrome) navidrome app gets IO error, what is interestng a navidrome's library build works on file metadata, but when a song is about to be fetched it fails.

I tried mount with 4.1 manually in navidrome running app and it started working, so I thought it would be nice to have an option to change the list of probes.

the hesitation to accept is reasonable. let me debug the problem further with the NFS server (which I also run as HA addon) maybe it somehow corrupts the connection

alexbelgium:
That would be great thanks - or else we could find how to improve the automatic code. It could for example test to write and read a file after mounting to verify if it works. This is how I set up the SMB system, and avoid needing to specify the version also. Thanks for your help in making the code better !

morgaroth:
found it
it was NFS server addon using buggy version of the NFS implementation, 4.2 was exposed, but buggy
upgraded one let navidrome connect properly

https://github.com/sam-ward/nfs-ganesha-addon/pull/4

hence closing this one, as you said it's unnecessary configurability

alexbelgium:
Glad it worked for you !

PR #3074, opened 2026-09-19
feat(calibre-web): optional Ingress login as the Home Assistant user

### LINKED ISSUE #3072: ✨ [REQUEST] calibre-web - Home Assistant login
### Checked

- [x] If a new addon, I have checked on  Google that such as addon doesn't already exists
- [x] If a new feature, I have searched in the discussions & issues that it was not already covered

### Which addon?

calibre-web

For calibre-web we have an ingress user variabel, which always logs in as that user in Home Assisant. When having multiple users in calibre, this is inconvenient. I would rather want to have calibre-web automatically log in as the Home Assistant user.

Remove ingress user, login as Home Assistant user. I discussed with an agent, who told me this would be possible. I can just not test it - maybe you see whether this has a chance to work?

[0001-feat-calibre-web-optional-login-through-Ingress-as-t.patch](https://github.com/user-attachments/files/32379452/0001-feat-calibre-web-optional-login-through-Ingress-as-t.patch)

alexbelgium:
Closes #3072.

## What changed
New opt-in option `login_with_ha_user` (bool, off by default). When enabled, the ingress nginx sets calibre-web's `X-WebAuth-User` header from the `X-Remote-User-Name` header the Supervisor adds to every ingress request (`supervisor/api/ingress.py` `_init_header`), instead of the fixed `ingress_user`. Each Home Assistant user then lands on the calibre-web account with the same name.

Not using the patch attached to the issue: it removes `ingress_user` and resets header login on upgrade, which changes behaviour for every current user. This PR is additive only.

## Existing users
With the option unset, `32-nginx.sh` produces exactly the same `ingress.conf` as before (`ingress_user`, default `admin`), and `80-configuration.sh` is untouched.

## Verified
- Local nginx with the generated `proxy_set_header X-WebAuth-User $http_x_remote_user_name;`: `X-Remote-User-Name: alice` reaches the backend as `X-WebAuth-User: alice`; no header sends nothing (calibre-web shows its login page); a client-supplied `X-WebAuth-User: admin` is dropped, not passed through.
- Supervisor source: it strips any client-supplied `X-Remote-User-*` headers before setting its own, so the username cannot be forged through ingress.
- `validate.sh --vs-master`: no new lint findings (the ShellCheck error in `services.d/nginx/finish` is pre-existing).

## Not verified
- End-to-end in a running add-on with multiple HA users (no dockerd here; CI is the build gate).
- HA users with no username (e.g. some non-owner setups) get no header and see the calibre-web login page — by design, but not exercised live.

## Rollback
Revert the `login_with_ha_user` block in `rootfs/etc/cont-init.d/32-nginx.sh`; with the option off it is already inert.

## Summary by CodeRabbit

- **New Features**
  - Added an optional setting to authenticate Ingress access using the Home Assistant username.
  - When enabled and a matching Calibre-Web account exists, users are signed in automatically; otherwise, the standard login page is displayed.
  - The setting is disabled by default, preserving existing login behavior.

- **Documentation**
  - Documented the new authentication option and its default value.

- **Chores**
  - Updated the add-on version to 0.6.27.5.


klassm:
Thank you so much ❤️

alexbelgium:
Happy to be able to improve life of other peoples :)

alexbelgium:
I mean, even if in a tiny bit ;)

PR #3079, opened 2026-09-20
fix(birdnet-pipy): ship the 0.8.9 stream supervisor and start icecast to match the upstream

Suncuss:
## Summary

BirdNET-PiPy **0.8.9** (released today, tag `v0.8.9`) replaces the shell-owned FFmpeg stream loops in `start-icecast.sh` with a Python stream supervisor (`deployment/audio/scripts/stream_supervisor.py`) that owns one publisher per audio source and reloads sources live. The script now ends with:

```
exec python3 /app/stream_supervisor.py
```

This Dockerfile copies only `start-icecast.sh`, so an image built from 0.8.9 has Icecast running but no publishers: the Live Feed shows "Stream dropped — reconnecting" and Settings reports the live stream as unavailable. (Upstream 0.8.9 also made the script idle instead of restart-looping when the supervisor is missing, so the add-on Log tab stays quiet — but there is still no stream.)

## Root cause

Two add-on specifics:

1. **The supervisor file is not in the image.** Only the shell script is copied.
2. **The icecast service runs under `gosu icecast2`, but the API service runs as root.** The API writes `user_settings.json` with mode 0600 into the root-owned data directory. The supervisor must read that file and write `streaming_status.json` beside it, which the `icecast2` account cannot do (reproduced: `PermissionError` on the status write). Icecast itself refuses to start as root, so simply dropping `gosu` is not enough.

## Fix

- Copy `deployment/audio/scripts/` **as a directory** to `/usr/local/lib/birdnet-audio/`, so the build succeeds on 0.8.8 (no supervisor file) and on 0.8.9+ alike.
- `services.d/icecast/run` picks the startup mode at boot:
  - supervisor present (0.8.9+): run as root. Upstream's `start-icecast.sh` detects root and adds `<changeowner>` so the Icecast daemon itself drops to `icecast2`, while the supervisor keeps root for the root-owned settings and status files.
  - supervisor absent (0.8.8): unchanged behaviour under `gosu icecast2`, log chown included.

Safe to merge before or after the updater bumps to 0.8.9. No version bump included: if 0.8.9 has already been built when this merges, a packaging bump (`0.8.9-1`) is needed to trigger a rebuild.

## Testing

- Upstream icecast image started as root with a root-owned 0600 `user_settings.json`: Icecast logged `Changed userid to 100`, the supervisor ran as root, connected the configured source, wrote `streaming_status.json`, and Icecast answered on 8888. Same result with a pre-multi-source (`rtsp_urls`) settings file.
- Missing-supervisor path (0.8.8-style tree): the service logs one error and idles with Icecast up.
- Our test add-on repo ([Suncuss/hassio-addon-birdnet-pipy](https://github.com/Suncuss/hassio-addon-birdnet-pipy)) publishes 0.8.9-dev4 with a byte-identical `icecast/run` and the same Dockerfile change for verification on a Home Assistant device.

https://claude.ai/code/session_01QhVJknwRMMhPFsuDNJRJGi

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Icecast service startup to support environments with an audio stream supervisor.
  - Added a fallback startup path for environments where the supervisor is unavailable.
  - Improved handling of audio service scripts and permissions during startup.
  - Ensured bundled startup scripts are executable in deployed builds.


alexbelgium:
Thanks!

